PR #7, opened 2021-05-25
Hotfix 6 timestamp

### LINKED ISSUE #6: ISO Timestamp issue
1. the documentation on GET syntax needs to be updated to include optional START/STOP
2. the syntax in lark requires ISO timestamp to have the milliseconds part, which should not be true

subbyte:
1. fix #6 
2. add `SORT` to doc

### DIFF
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -675,6 +675,36 @@ Examples
     procs = GET process FROM stixshifter://edrA WHERE [process:parent_ref.name = 'bash']
     DISP procs ATTR pid, name, command_line
 
+SORT
+----
+
+The command ``SORT`` is an *inspection* hunt step to reorder of entities in a
+Kestrel variable and output the same set of entities with new order to a new
+variable.
+
+Syntax
+^^^^^^
+::
+
+    newvar = SORT varx BY stixpath [ASC|DESC]
+
+- The ``stixpath`` can be a full STIX path like ``process.attribute`` or just
+  an attribute name like ``pid`` if ``varx`` is ``process``.
+
+- By default, data will be sorted by descending order. The user can specify the
+  direction explicitly such as ``ASC``: ascending order.
+
+Examples
+^^^^^^^^
+::
+
+    # get network traffic and sort them by their destination port
+    nt = GET network-traffic FROM stixshifter://idsX WHERE [network-traffic:dst_ref_value = '1.2.3.4']
+    ntx = SORT nt BY dst_port ASC
+
+    # display all destination port and now it is easy to check important ports
+    DISP ntx ATTR dst_port
+
 GROUP
 -----
 
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -445,7 +445,7 @@ Syntax
 ^^^^^^
 ::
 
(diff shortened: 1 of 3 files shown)


docs/language.rst
```diff
@@ -257,7 +257,7 @@ Syntax
 ^^^^^^
 ::
 
-    returned_variable = GET returned_entity_type FROM entity_pool WHERE stix_pattern
+    returned_variable = GET returned_entity_type FROM entity_pool WHERE stix_pattern [START t'timestamp' STOP t'timestamp']
```
pcoccoli: Do the square brackets here mean "optional"?  It's a bit confusing since the START/STOP qualifier is part of the `stix_pattern`.

docs/language.rst
```diff
@@ -657,6 +675,36 @@ Examples
...
+Kestrel variable and output the same set of entities with new order to a new
+variable.
+
+Syntax
+^^^^^^
+::
+
+    newvar = SORT varx BY stixpath [ASC|DESC]
+
+- The ``stixpath`` can be a full STIX path like ``process.attribute`` or just
```
pcoccoli: This should be `process:attribute` right?  A STIX object path is defined as <SCO type>:<property>

docs/tutorial.rst
```diff
@@ -327,7 +336,10 @@ We can find connected entities easily in Kestrel, e.g., child processes created
 of processes, network traffic created by processes, files loaded by processes,
 users who own the processes. To do so, use the ``FIND`` command with a
 previously created Kestrel variable, which stores a list of entities from which
-to find connected entities. Read more in :doc:`language`.
+to find connected entities. Note that not all data sources have relation data,
+and not all STIX-Shifter are mature enough to translate relation data. The data
```
pcoccoli: Should read "not all STIX-Shifter connector modules are mature enough"

pcoccoli (changes requested):

docs/language.rst
subbyte: Yes. This is a hard choice and a hack. For new users, the first query usually needs to have the time range, but most new users even don't know `stix_pattern` and how it structures. When seeing an example of `[process:name = 'xyz' and process:pid = '123']`, they may assume STIX pattern is in the square bracket (happened today). I had the description of using time range, but it was burred in the usage of `stix_pattern`, and it takes too long to understand STIX for new users, and also why START/STOP need to appear together. So I am thinking to hack it (put it explicit as an option). I know this break stix pattern, which can have multiple time ranges, one for each observation expression. Would like to hear your further suggestion.

docs/language.rst
subbyte: Good one. Doc fixed. And we need to do a better lark file to rule out impossible cases in Q3.

docs/tutorial.rst
subbyte: Thanks for the catch! Fixed.

docs/language.rst
subbyte: Decided to remove the confusing part from syntax and try to explain in doc better.

pcoccoli (approved):

Issue #32, opened 2021-06-04
codegen upgrade: use firepit to merge variables


subbyte:
**Is your feature request related to a problem? Please describe.**
There are several shortcuts in `codegen/commands.py` currently using `or_pattern()`. This is not performance friendly.

**Describe the solution you'd like**
Need to use the merge API from `firepit` to implement it.

subbyte:
`or_pattern()` completed removed in #253

PR #33, opened 2021-06-05
implement entity id attr pick-up mech and fix #31

### LINKED ISSUE #31: process pid column missing
**Describe the bug**
PostgreSQL backend (if used) will give an error when `pid` in process does not exist.

**Details of the bug**
- What is the hunt flow/script you are executing?
`test_new_with_list_of_strings()` in `tests/test_command_new.py`

- What is the command that failed?
The `NEW` command

- What is the error message?
No `pid` column in database.

**Expected behavior**
Should work.

subbyte:
The PR fully addresses #31 and partially addresses #32.

1. implement a new function `get_entity_id_attribute()` in `src/kestrel/codegen/relations.py` to compute the appropriate attribute used as identifier attribute for entities.
2. update `src/kestrel/codegen/commands.py` to use `get_entity_id_attribute()`.
3. replace the `or_pattern()` for post-prefetch merge in `src/kestrel/codegen/commands.py` with `firepit.merge()` (partially address #32).
4. update `get_variable_entity_count()` in `src/kestrel/codegen/summary.py` to use `get_entity_id_attribute()`.
5. update `_get_variable_query_ids()` in `src/kestrel/codegen/summary.py` since merged variable in `firepit` do not have `__membership` records.
6. update `gen_variable_summary()` in `src/kestrel/codegen/summary.py` to only give cached records when there is a data source query.

### DIFF
--- a/.github/workflows/code-style.yml
+++ b/.github/workflows/code-style.yml
@@ -32,4 +32,4 @@ jobs:
           python -m pip install black
           python -m pip install .
       - name: Code style check
-        run: python -m black --check src/
+        run: black --check src/
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -23,5 +23,5 @@ jobs:
           TWINE_USERNAME: __token__
           TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
         run: |
-          python -m build --sdist --wheel --outdir dist/ .
-          python -m twine upload dist/*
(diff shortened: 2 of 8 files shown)


pcoccoli (approved):
Looks OK, though ideally we should be able to use multiple attributes for object identity (e.g. for a process, use *both* pid and name).

subbyte:
> Looks OK, though ideally we should be able to use multiple attributes for object identity (e.g. for a process, use _both_ pid and name).

Good point to make the function `get_entity_id_attribute()` more comprehensive regarding logic in the future.

PR #47, opened 2021-06-21
Doc update

jmcasava:
Reviewed and editing for style and grammar


subbyte (changes requested):
Thanks very much for your contribution, Jill! Looks good except that the `.DS_Store` file should not be committed. Please remove that file in the PR.

PR #64, opened 2021-06-29
Infer scheme if omitted and only 1 is available

pcoccoli:
If you only have 1 scheme (datasource or analytics), it would be nice if you didn't have to specify it. 
In the future, we could also add a "default" scheme to the config file, and use that if the scheme is omitted.
Another future possibility is to walk the list of all choices for each scheme and use the first match.


subbyte:
Both are very good suggestions. However, there is a small issue on the second choice: in case there are two data sources that have the same name but in different schema, the first match will work, yet it is not obvious to the user how we sort the schema when traversing the data sources in schemas. The first choice is more obvious in my mind: when the user/system configure Kestrel, it is clear what schema by default will be used.

subbyte (approved):

PR #98, opened 2021-08-11
Removed keyword argument "force" from logging.basicConfig.

### LINKED ISSUE #97: ValueError: Unrecognised argument(s): force
**Describe the bug**
ValueError when running Hello World Hunt using Python 3.6.9. Installed using pip install kestrel-lang.

**Details of the bug**
- What is the hunt flow/script you are executing?
Hello World Hunt from readme
- What is the command that failed?
```
$ kestrel helloworld.hf
```
- What is the error message?
```
$ kestrel helloworld.hf
  File "/usr/local/bin/kestrel", line 8, in <module>
    runpy.run_module('kestrel', run_name='__main__')
  File "/usr/lib/python3.6/runpy.py", line 208, in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.6/dist-packages/kestrel/__main__.py", line 49, in <module>
    logging_setup(None, args.verbose, args.debug)
  File "/usr/local/lib/python3.6/dist-packages/kestrel/__main__.py", line 33, in logging_setup
    force=True,
  File "/usr/lib/python3.6/logging/__init__.py", line 1829, in basicConfig
    raise ValueError('Unrecognised argument(s): %s' % keys)
ValueError: Unrecognised argument(s): force
```

**To Reproduce**
1. pip install --upgrade pip setuptools wheel
2. pip install kestrel-lang
3. kestrel helloworld.hf

**Expected behavior**
Output from hunt flow.

**Environment (please complete the following information):**
- OS: Ubuntu 18.04
- Python version: Python 3.6.9
- Python install environment: Python virtual environment
- STIX-Shifter version: 3.5.0

kinzhong:
This pull request fixes https://github.com/opencybersecurityalliance/kestrel-lang/issues/97.

### DIFF
--- a/src/kestrel/__main__.py
+++ b/src/kestrel/__main__.py
@@ -37,6 +37,7 @@ def logging_setup(session, verbose_mode, debug_mode):
         handlers=log_handlers,
     )
 
+
 if __name__ == "__main__":
 
     parser = argparse.ArgumentParser(description="Kestrel Interpreter")
--- a/src/kestrel/__main__.py
+++ b/src/kestrel/__main__.py
@@ -25,13 +25,17 @@ def logging_setup(session, verbose_mode, debug_mode):
         log_handlers = [log_console, log_file] if verbose_mode else [log_file]
     else:
         log_handlers = [log_console]
+    # remove existing handlers attached to root logger
+    root_logger = logging.getLogger()
+    for h in root_logger.handlers[:]:
+        root_logger.removeHandler(h)
+        h.close()
     logging.basicConfig(
         format=log_format,
         datefmt="%H:%M:%S",
         level=logging.DEBUG if debug_mode else logging.INFO,
         handlers=log_handlers,
     )
-
 
 if __name__ == "__main__":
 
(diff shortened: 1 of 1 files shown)


subbyte (commented):
Thanks for the PR, @kinzhong !

We need to do a little more than just removing the parameter. The function `logging_setup()` is called twice in `__main__.py` before and after the session starts with logging file created. The expected behavior is to output log to stdout/stderr before the creation of the logging file and then into the logging file (may plus the stdout/stderr depending on `verbose_mode`). However, the root logger created by `basicConfig` will not be completely replaced by the new one unless `force` is used. That is why I say we need to manually remove the root loggers such as stderr if we do not use `force=true` --- basically implement the functionality of `force` for `logging` (when the earlier versions does not have the parameter).

kinzhong:
I see. I have added that functionality. Please take a look @subbyte. Also, I apologize in advance if there are mistakes. It's my first time dealing with Python logging!

subbyte (changes requested):
Great! The last change needed is to black the code. Just `pip install black` and `black src` which should take care of the file you edited.

subbyte (approved):

PR #134, opened 2021-10-08
Execution tracking and diagram

raymundl:
Tracking all executions and their output variables, including failed attempts, which eventually visualizes [Hunt Flow](https://kestrel.readthedocs.io/en/latest/_images/huntflow.png).
It also tells the hunter the current state of hunting, by visualizing all variable versions, and how they are generated.
10  DISP
In  5  DISP

### DIFF
--- a/docs/language.rst
+++ b/docs/language.rst
@@ -723,12 +723,25 @@ Syntax
 ^^^^^^
 ::
 
-    aggr_var = GROUP varx BY attribute
+    aggr_var = GROUP varx BY attr1, attr2... [WITH aggr_fun(attr3) [AS alias], ...]
 
-- Currently, the command only supports aggregation by one attribute.
+- If no aggregation functions are specified, they will be chosen
+  automatically.  In that case, attributes of the returned entities
+  are decorated with a prefix ``unique_`` such as ``unique_pid``
+  instead of ``pid``.
 
-- Attributes of the returned entities are decorated with a prefix ``unique_``
-  such as ``unique_pid`` instead of ``pid``.
+- When aggregations are specified without ``alias``, aggregated
+  attributes will be prefixed with the aggregation function such as
+  ``min_first_observed``.
+
+- Support aggregation functions:
+
+  - ``MIN``: minimum value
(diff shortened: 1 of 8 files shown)


subbyte (commented):
Thanks much for this exciting PR that implements #140 towards #141, Raymund! Considering #141, let's think bigger and upgrade this PR.
1. Not overuse `DISP _` (`_` is a special variable in Kestrel named `default_variable`, which is defined in `config/kestrel.yaml` and always set to the last variable maintained in `session.py`).
2. Let's create a new command for this PR to implement #140 (later will be used in #141). Maybe `SUMMARIZE FLOW` for #140 and `SUMMARIZE HUNT` for #141? Your suggestion?
3. Let's use a more expressive name for `config/execution_tracking_diagram.txt`. Maybe `config/huntflow_diagram_template.html`?
4. Why do we need to hardcode the diagram template in `codegen/commands.py` again besides the loading from config folder?
5. When does a user need the new parameter `blockname` in `execution()` in `session.py`? How can a user provide it?
6. The variable dependency mining part in the new function `track_execution()` is already realized in `VarStruct.dep_vars` in `symboltable.py`. I planned `dep_vars` exactly for your PR :-) In other words, there is no need to set `track_execution = true/false` in `config/kestrel.toml`---variable dependency tracking is always enabled. Please make use of `VarStruct.dep_vars` and maybe we can move `track_execution()` largely from `session.py` to `codegen/commands.py` under the execution of the new `SUMMARIZE` command since the runtime can trace back the dependency graph linking VarStruct at any time---it is good to do it only once when the user execute `SUMMARIZE` to summarize the hunt flow.

raymundl:
1. Ok
2. The command names sound good to me, let me work on it
3. Ok
4. Good point. I added config later, and forgot I can get rid of inline version. Let me revise.
5. Currently needed only for IBM products, which use a "step name" as abstraction of the whole block of statements. I'm not sure if Jupyter users need it. Therefore it's default None.
6. Cool. Let me check out dep_vars and revise my PR.

PR #152, opened 2021-12-15
dollar sign display correction in Jupyter

subbyte:
fixed https://github.com/opencybersecurityalliance/kestrel-jupyter/issues/6


pcoccoli (approved):
I can't think of any negative effects.

PR #161, opened 2022-01-26
Feature 160 config merge

### LINKED ISSUE #116: Errors when kestrel.toml goes out-of-date
**Describe the bug**
Users can have a custom kestrel.toml with their own config.  When the code is updated to use new config values, the user gets errors like KeyError, etc.

**Details of the bug**
I pulled the latest code and ran pytest; most tests fail. 
I also ran `kestrel ...` with a test file and got this:
```
  File "/home/pcoccoli/.pyenv/versions/gh392/bin/kestrel", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/pcoccoli/github/kestrel-lang/bin/kestrel", line 8, in <module>
    runpy.run_module('kestrel', run_name='__main__')
  File "/home/pcoccoli/.pyenv/versions/3.9.2/lib/python3.9/runpy.py", line 213, in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
  File "/home/pcoccoli/.pyenv/versions/3.9.2/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/pcoccoli/github/kestrel-lang/src/kestrel/__main__.py", line 54, in <module>
    with Session(debug_mode=args.debug) as session:
  File "/home/pcoccoli/github/kestrel-lang/src/kestrel/session.py", line 179, in __init__
    if debug_mode or os.getenv(self.config["debug"]["env_var"], False)
KeyError: 'debug'
```
In this case, a new "debug" section was added between versions 1.1.0 and 1.1.1.

**Expected behavior**
Either kestrel just uses hardcoded defaults for config values missing from the TOML file, or an error explaining that the TOML file needs to be updated.

### LINKED ISSUE #160: Normalize configuration to YAML
Currently there are two configurations: main (TOML) and stix-shifter (YAML)

Todo:
- normalize them to YAML.
- reuse functions between them.
- write docs on configuration.

And due to the complexity that Python packages can be installed by OS package manager, e.g., `apt-get`, `dnf`, `pacman`, or can be installed by directly invoking `pip` by users in standard/virtualenv/conda environments, the configuration file will be written to different places.

Existing logic in v1.1.6 uses a list of possible config file places to load config files, and assign priorities of config files based on their position in the list (the later can override the former). This may cause an issue when the user does not remember which way he installed the package with different versions (different config files) and installed it twice using `dnf` and `pip` in a virtualenv.

To avoid confusion and make it easy to users to config. The following logic will be provided in this issue/PR:
1. default config is not in a file, but loaded from a Python module directly
2. only one folder is checked for user-defined config: `~/.config/kestrel/` to override default configs
3. docs will expose default config and needs to be updated when the default config changes

subbyte:
redo Kestrel config for both main Kestrel config and stix-shifter config/profiles and fix #160 
- YAML config for both, reusable functions
- default config not handled by pip, and fix #116
- enable stateful datasource interface config, and add `config` to interface API functions
- separate Kestrel and stix-shifter log level with env var `KESTREL_STIXSHIFTER_DEBUG`
- doc for Kestrel main config
- doc for stix-shifter interface config


pcoccoli (approved):
Looks good to me

PR #169, opened 2022-02-09
Python analytics interface

subbyte:


docs/debug.rst
```diff
@@ -0,0 +1,72 @@
...
+
+    .. code-block:: python
+
+        _logger.debug(something_you_want_to_log)
+
+#. Install your local Kestrel build:
+
+    .. code-block:: console
+
+        $ pip install .
```
pcoccoli: I recommend `pip install -e .` here

src/kestrel_datasource_stixshifter/interface.py
```diff
@@ -109,14 +114,26 @@
...
     except:
-        raise DataSourceError(
-            f'STIX shifter connector for "{connector_name}" is not installed',
-            "please install the corresponding STIX shifter connector Python package.",
+        package_name = "stix-shifter-modules-" + connector_name.replace("_", "-")
+        _logger.info(
+            f'installing missing stix-shifter connector "{connector_name}" '
+            f'with possible Python package "{package_name}"'
         )
+        subprocess.check_call([sys.executable, "-m", "pip", "install", package_name])
```
pcoccoli: I'm not convinced this is a good idea...

pcoccoli (approved):
Mostly looks good to me; I added a couple of comments.

src/kestrel_datasource_stixshifter/interface.py
subbyte: This is a usability vs. security balance. It makes using a stix-shifter connector much more convenient (I couldn't find the mapping between connector name to package, so as many users who may not know stix-shifter well), but open a new attack surface where attackers may upload their own stix-shifter modules to pypi.

subbyte: Will revert to the method of giving error and suggestions.

subbyte: Use PyPI to verify the stix-shifter connector package is genuine before install.

PR #182, opened 2022-03-04
STIX bundle optimizations

pcoccoli:
1. Download and save bundle files to avoid multiple downloads
2. Pre-compile STIX pattern before matching

### DIFF
--- a/src/kestrel_datasource_stixbundle/interface.py
+++ b/src/kestrel_datasource_stixbundle/interface.py
@@ -1,7 +1,10 @@
 import json
+import logging
 import re
 import uuid
 import pathlib
+from datetime import datetime, timezone
+from dateutil import parser
 import requests
 
 from stix2matcher.matcher import match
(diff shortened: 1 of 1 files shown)


src/kestrel_datasource_stixbundle/interface.py
```diff
@@ -43,41 +48,78 @@ def query(uri, pattern, session_id=None, config=None):
...
         data_paths = data_paths.split(",")
         pattern = fixup_pattern(pattern)
+        compiled_pattern = Pattern(pattern)
 
         ingestdir = _make_query_dir(uri)
         bundles = []
         for i, data_path in enumerate(data_paths):
             data_path_striped = "".join(filter(str.isalnum, data_path))
-            ingestfile = ingestdir / f"{i}_{data_path_striped}.json"
+            bundlefile = ingestdir / f"{data_path_striped}.json"
```
subbyte, line 57: Each time `query()` is called, there should be a random `ingestdir`, which may make caching not friendly. Maybe we'd like to put all downloaded bundles not in `ingestdir/...` but at a fixed name dir, e.g., like `stixbundles` under the runtime dir. We may need to append the modified time to the file name explicitly (download time if that not exist) so the same name json from different sites will not collide.

pcoccoli: I think that's a further optimization that we could (and probably should) do, but I don't have time right now.

subbyte: I think my concern may not exist given we are using `uuid5()` to generate `ingestdir`. Please ignore it.

subbyte (approved):
smart idea to check caching with modified time. Looks good!

PR #183, opened 2022-03-07
Skip prefetch when it won't return anything new

pcoccoli:
Don't do prefetch for get when the STIX pattern in GET is "direct" (i.e. the return type is the only type used in the pattern) and the properties used are "ID" attributes.


subbyte (approved):
Looks good!

PR #199, opened 2022-04-15
Improve do_complete

pcoccoli:
- add unit tests for auto-completion
- update grammar to separate commands that yield a variable and those and that don't
- add some special case handling in do_complete

### DIFF
--- a/src/kestrel/session.py
+++ b/src/kestrel/session.py
@@ -368,7 +368,9 @@ class Session(AbstractContextManager):
             A list of suggested strings to complete the code.
         """
         prefix = code[:cursor_pos]
-        last_word = prefix.split(" ")[-1]
+        words = prefix.split(" ")
+        last_word = words[-1]
+        _logger.debug('code="%s" prefix="%s" last_word="%s"', code, prefix, last_word)
 
         if "START" in prefix or "STOP" in prefix:
             return self._get_complete_timestamp(last_word)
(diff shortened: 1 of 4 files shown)


src/kestrel/session.py
```diff
@@ -418,12 +431,23 @@ def do_complete(self, code, cursor_pos):
...
+                            tmp.extend(get_entity_types())
+                        else:
+                            tmp.extend(all_relations)
                     elif token == "REVERSED":
                         tmp.append("BY")
-                        varnames = self.get_variable_names()
-                        if last_word not in varnames:
+                        prev_word = words[-2] if len(words) >= 2 else ""
+                        _logger.debug("prev_word = %s", prev_word)
+                        if prev_word in all_relations:
```
subbyte, line 446: I think we can simplify `elif token == "REVERSED":` for only the case of `FIND`. We use keyword `"by"i` in `GROUP` and `SORT`, but not under `REVERSED`. I like the way we currently do the syntax, which differentiates generic `"by"i` from the special use of it as `REVERSED`. My guess is we do not need to handle those cases here.

subbyte (commented):
Timely upgrade! Just one place I don't fully get the logic. We may get it a little simpler there.

src/kestrel/session.py
pcoccoli: When `token == "REVERSED"` we don't know if it's a `FIND`, `GROUP`, or `SORT`.  That's why I look back to the previous word and check if it's a relation.

subbyte: I thought the parser will only suggest `REVERSED` in `FIND`, `"by"i` in `GROUP` and `SORT` are not under `REVERSED` in the Lark syntax.

pcoccoli: Surprisingly, that's not the case.  You can see that when running the test cases.

pcoccoli: For example, the test case `grps = GROUP conns ` hits the `REVERSED` case, even though we use the literal `"by"i` in the rule for `group`.

subbyte: Interesting. Just did a Lark test, and this looks like a Lark logic that tries to use a terminal if exist. Also found a possible fix to tell Lark to differentiate them. Let me have a quick try to simplify it.

Issue #200, opened 2022-04-15
No warning given for invalid entity type


pcoccoli:
**Is your feature request related to a problem? Please describe.**
There's no error checking done for entity type (e.g. in a `GET` command).  While it's an "open vocabulary" (i.e. data sources can use custom object types), if we don't find any entities we should consider showing some kind of warning.

**Describe the solution you'd like**
Perhaps a warning or something to alert the user that they may have mis-typed the entity type.  Custom objects *should* start with `x-` so if there are 0 matches AND the name doesn't start with `x-` AND the name is not a known entity type, show some kind of warning.

**Describe alternatives you've considered**
I can't think of any other options.

**Additional context**
Reported by a user.

PR #209, opened 2022-04-19
DisplayWarning

pcoccoli:
Display a warning to the user when they do a `GET` for an unknown entity (SCO) type


subbyte (approved):

pcoccoli:
Fixes #200

PR #211, opened 2022-04-19
fix two issues for complete FIND command

subbyte:
1. the terminal `"by"i` is now differentiated between `FIND` and `SORT`/`GROUP`
2. now variables such as `conns` are not in the completion of `FIND process c`


pcoccoli (approved):
Nice!

PR #216, opened 2022-04-21
use tmp dir for testing SAVE command

subbyte:


pcoccoli (approved):
Looks fine to me.  I don't really understand the codecov output - why would this decrease coverage?

PR #217, opened 2022-04-21
Deref mixed ips

pcoccoli:


pcoccoli:
The github actions failed because firepit 2.0.1 just got published - it seems to take a little while before the new version can be found.

subbyte (approved):

Issue #226, opened 2022-05-10
dynamic stix-shifter config loading


subbyte:
**Is your feature request related to a problem? Please describe.**
When using binder, a user needs a way to input stix-shifter config so he/she can test hunts with data in his/her environments.

**Describe the solution you'd like**
The simplest solution:
1. load stix-shifter config file each time a list/execute function is called in the interface
2. figure out a way for users to edit `stixshifter.yaml`.
  a. use the current directory (cwd) as the stix-shifter config dir in binder env.
  b. prepare a `stixshifter.yaml` file in the current dir for users to edit, which is linked to the default config file.

The Jupyter Notebook binder launched already has a _text editor_, which can be used to create/edit `stixshifter.yaml` file.

**Describe alternatives you've considered**
`CONFIG` command. However, it is unclear how to handle password in the UI. We may need [Jupyter Widgets](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Basics.html) to obtain asterisked password.

subbyte:
fixed with #227 and https://github.com/opencybersecurityalliance/kestrel-huntbook/commit/fa0ecb5d09d27f86d81a2bfe36b75bc114305661

PR #227, opened 2022-05-11
Feature 226 stixshifter config loading

subbyte:


subbyte:
fix #226 on the code side. Need to adjust binder rc file.

PR #228, opened 2022-05-13
Fix PR 225

### LINKED ISSUE #224: AttributeError with timestamped grouped variable
**Describe the bug**
You shouldn't be able to use the `TIMESTAMPED` transform with grouped/aggregate variables, but we don't guard against it currently.

**Details of the bug**
```
  File "/home/pcoccoli/github/kestrel-lang/src/kestrel/codegen/commands.py", line 216, in disp
    qry = _build_query(session.store, entity_table, qry, stmt)
  File "/home/pcoccoli/github/kestrel-lang/src/kestrel/codegen/commands.py", line 648, in _build_query
    _set_projection(store, entity_table, qry, cols)
  File "/home/pcoccoli/github/kestrel-lang/src/kestrel/codegen/commands.py", line 633, in _set_projection
    for p in proj.cols:
AttributeError: 'NoneType' object has no attribute 'cols'
```

**To Reproduce**
```
conns = GET network-traffic
        FROM file:///home/pcoccoli/github/kestrel-lang/tests/test_bundle.json
	WHERE [network-traffic:dst_ref.value NOT ISSUBSET '192.168.0.0/16']

grp_conns = GROUP conns BY dst_ref.value WITH COUNT(dst_ref) AS count

DISP TIMESTAMPED(grp_conns)
```

**Expected behavior**
The user should get an error message stating that you can't use `TIMESTAMPED` with grouped/aggregate vars.

**Environment (please complete the following information):**
Kestrel v1.3.2

pcoccoli:
#255 introduced a bug where subsequent `GET` would return no results.  This PR fixes that, at the cost of mostly undoing the `process` de-duplication #225 was meant to address.
Also included in this PR:
- upgrade to firepit 2.0.3 for bugfixes
- fix #224 AttributeError with timestamped grouped variable

### DIFF
--- a/tests/test_command_find.py
+++ b/tests/test_command_find.py
@@ -81,10 +81,10 @@ files = FIND file LINKED procs
         s.execute(stmt)
         procs = s.get_variable("procs")
         print(json.dumps(procs, indent=4))
-        assert len(procs) == 7
+        assert len(procs) == 7 * 2
         files = s.get_variable("files")
         print(json.dumps(files, indent=4))
-        assert len(files) == 2
+        assert len(files) == 2 * 3  # FIXME: why * 3?!
 
 
 def test_find_file_loaded_by_process(proc_bundle_file):
@@ -98,7 +98,7 @@ files = FIND file LOADED BY procs
         s.execute(stmt)
         procs = s.get_variable("procs")
         print(json.dumps(procs, indent=4))
-        assert len(procs) == 7
+        assert len(procs) == 7 * 2
         files = s.get_variable("files")
         print(json.dumps(files, indent=4))
         assert len(files) == 1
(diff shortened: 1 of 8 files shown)


tests/test_command_find.py
```diff
@@ -81,10 +81,10 @@ def test_find_file_linked_to_process(proc_bundle_file):
         s.execute(stmt)
         procs = s.get_variable("procs")
         print(json.dumps(procs, indent=4))
-        assert len(procs) == 7
+        assert len(procs) == 7 * 2
         files = s.get_variable("files")
         print(json.dumps(files, indent=4))
-        assert len(files) == 2
+        assert len(files) == 2 * 3  # FIXME: why * 3?!
```
subbyte, line 87: It could be due to the fact that in the input bundle `doctored-1k.json`, the processes with command line `wmic%` has `binary_ref` (very good and complete data source), and I guess the `[process:command_line LIKE 'wmic%']` pattern cached the files once, followed by prefetch of `procs` cached them twice, then the prefetch of `FIND` based on file names cached them the third time.

subbyte (approved):

PR #234, opened 2022-06-15
group by 'binned' attributes

pcoccoli:


src/kestrel/syntax/kestrel.lark
```diff
@@ -117,6 +117,14 @@ variables: VARIABLE ("," VARIABLE)*
 
 path_list: STIXPATH ("," STIXPATH)* -> valuelist
```
subbyte: Shall we remove `path_list` from the syntax?

src/kestrel/syntax/kestrel.lark
```diff
@@ -117,6 +117,14 @@ variables: VARIABLE ("," VARIABLE)*
...
 
+grp_spec: grp_expr ("," grp_expr)*
+
+grp_expr: STIXPATH
+        | bin_func
+
+bin_func: "BIN"i "(" STIXPATH "," NUMBER timeunit? ")"
```
subbyte, line 123: It is fine for the first version. In the future, we may improve it to be either:
- temporal attributes (maybe a subset of STIXPATH) with timeunit
- generic attribute with some other way to define binning unit

src/kestrel/syntax/kestrel.lark
pcoccoli: Done.

subbyte (approved):

PR #239, opened 2022-07-07
upgrade process identification to fix #238

### LINKED ISSUE #238: add ppid in fine-grained process filtering
In the following case:
```
p = GET process ...
pp = FIND process CREATED p
```
If the first command only returns STIX observations with `parent_ref.pid` but no `parent_ref.name`, then the second command will deref the parent process without names. In the current `fine_grained_relational_process_filtering()` logic, if the prefetched process records (for the parent) have name, while the deref records do not, it will be treated as a name-changing case, and possibly filtered out.

We haven't consider using the ppid information if available for process identification. This is the time to add this logic to not throw away possible matches when process name is not in data but ppid is.

subbyte:


src/kestrel/codegen/relations.py
```diff
@@ -230,52 +230,122 @@ def _generate_paramstix_comparison_expressions(
...
-    query_ref = Query(
-        [
-            Table(local_var.entity_table),
-            Join("__contains", "id", "=", "target_ref"),
-            Join("observed-data", "source_ref", "=", "id"),
-            Projection(["pid", "name", "first_observed", "last_observed"]),
-        ]
+    ref_processes = _query_process_with_time_and_ppid(store, local_var.entity_table)
+
+    fil_processes = _query_process_with_time_and_ppid(store, prefetch_entity_table)
```
pcoccoli: What are `ref_processes` and `fil_processes`?

subbyte: the logic is a little complex, comments added.

pcoccoli (approved):

PR #244, opened 2022-07-25
Support for automatically converting input files to STIX in stixbundle datasource module

pcoccoli:
Uses firepit's woodchipper if the local or downloaded file isn't STIX

### DIFF
--- a/src/kestrel_datasource_stixbundle/interface.py
+++ b/src/kestrel_datasource_stixbundle/interface.py
@@ -5,11 +5,12 @@ bundles locally or remotely.
 
 import json
 import logging
+import os
 import re
 import uuid
 import pathlib
 import shutil
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from dateutil import parser
 import requests
 
@@ -85,10 +86,6 @@ class StixBundleInterface(AbstractDataSourceInterface):
 
         bundles = []
         for i, data_path in enumerate(data_paths):
-
-            data_path_stripped = "".join(filter(str.isalnum, data_path))
-            ingestfile = ingestdir / f"{i}.json"
-
             _logger.debug(f"requesting data from path: {data_path}")
 
             if scheme == "file":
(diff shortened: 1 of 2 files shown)


src/kestrel_datasource_stixbundle/interface.py
```diff
@@ -90,20 +92,24 @@ def query(uri, pattern, session_id=None, config=None):
...
                     with open(data_path, "r") as f:
                         bundle_in = json.load(f)
                 except Exception:
                     _clean_ingestdir_and_raise_error(ingestdir, uri)
 
             elif scheme == "http" or scheme == "https":
-                bundlefile = downloaddir / f"{data_path_stripped}.json"
-                data_uri = f"{scheme}://{data_path}"
+                data_path, dot, extension = data_path.rpartition(".")
+                data_path_stripped = "".join(filter(str.isalnum, data_path))
```
subbyte, line 102: duplicate with line 89

src/kestrel_datasource_stixbundle/interface.py
```diff
@@ -90,20 +92,24 @@ def query(uri, pattern, session_id=None, config=None):
...
+                rawfile = downloaddir / f"{data_path_stripped}"
+                if extension:
+                    rawfile = rawfile.with_suffix(f".{extension}")
```
subbyte: Not sure why replacing extension with itself. Maybe I miss some special cases.

src/kestrel_datasource_stixbundle/interface.py
```diff
@@ -90,20 +92,24 @@ def query(uri, pattern, session_id=None, config=None):
...
                 try:
                     with open(data_path, "r") as f:
                         bundle_in = json.load(f)
                 except Exception:
                     _clean_ingestdir_and_raise_error(ingestdir, uri)
 
             elif scheme == "http" or scheme == "https":
-                bundlefile = downloaddir / f"{data_path_stripped}.json"
-                data_uri = f"{scheme}://{data_path}"
+                data_path, dot, extension = data_path.rpartition(".")
```
subbyte: What happen if the URL does not have extension like `www.sample.com/stixbundle` (but it is a JSON file)? It will give out `data_path, dot, extension = ('www.sample', '.', 'com/stixbundle')`

src/kestrel_datasource_stixbundle/interface.py
```diff
@@ -90,20 +92,24 @@ def query(uri, pattern, session_id=None, config=None):
...
+                data_path_stripped = "".join(filter(str.isalnum, data_path))
+                rawfile = downloaddir / f"{data_path_stripped}"
+                if extension:
+                    rawfile = rawfile.with_suffix(f".{extension}")
+                data_uri = f"{scheme}://{data_path}{dot}{extension}"
```
subbyte: Maybe a more robust way is the previous `data_uri = f"{scheme}://{data_path}"`, since the `rpartition()` may not work as expected.

subbyte (commented):
Thanks, Paul! Some minor comments that may help improve the flow.

src/kestrel_datasource_stixbundle/interface.py
pcoccoli: Fixed

src/kestrel_datasource_stixbundle/interface.py
pcoccoli: Replace with `os.path.splitext()`

subbyte (approved):

Issue #248, opened 2022-07-29
current directory for stixshifter config does not work


subbyte:
In the binder service, we let users to provide `stixshifter.yaml` at the current dir.

This does not work when executing a command, only works for auto-completion. This is due to the switch to tmp dir when executing a Kestrel command.

PR #261, opened 2022-10-26
relative path support with unit tests

subbyte:


subbyte:
fix #248 and beyond (3 cases of relative path resolution)

PR #275, opened 2023-01-18
Restrict assign and merge command syntax

subbyte:
The existing `assign` and `merge` commands allow omitted result variable, so the following huntflow is currently legal:
```
DISP var1 var2 var3 var4
```
Which basically is
```
DISP var1
_ = var2
_ = var3
_ = var4
```
This is not meaningful, which may confuse users, and it fails existing auto-completion.

This PR restricts `assign` and `merge` commands so they need to have a result variable. This is a preparation step for auto-completion upgrade. Now a variable alone is not a complete Kestrel statement anymore, so `varx` will expect `=`.

### DIFF
--- a/src/kestrel/session.py
+++ b/src/kestrel/session.py
@@ -469,7 +469,10 @@ class Session(AbstractContextManager):
                     elif token.startswith("__ANON"):
                         continue
                     elif token == "EQUAL":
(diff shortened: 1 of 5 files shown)


pcoccoli (approved):
Looks good

Issue #264, opened 2022-11-11
autocomplete function doesn't behave correctly


vereimyst:
**Description:** 
The [autocompletion function](https://github.com/opencybersecurityalliance/kestrel-lang/blob/4f1fa4262a08bbadb42718962fa5553123ee0adf/src/kestrel/session.py#L350) doesn't properly address partially complete fields correctly. Fields with the same starter characters as commands also return commands matching `last_word` as suggestions (which they should not be doing). No error messages are outputted, but the behavior does not match what is expected. Examples included below for clarity. A possible solution mentioned is completely revamping the logic of the `docomplete()` function, looking at the parsing portion of the existing code in particular.

**Environment:**
I verified that the variable autocompletion error appears in the Tutorial Huntbook environment, so I believe any Kestrel runtime environment through Jupyter Notebook should display this issue.
For attribute autocompletion, I was running my Kestrel environment from Jupyter Notebook locally hosted through a Python 3 virtual environment on Windows 11 WSL Ubuntu 20.04.5 LTS. (I have no idea if the grammar there was correct, sorry...)

**Details & How to Reproduce:** 
Huntflow taken directly from the [Kestrel tutorial](https://mybinder.org/v2/gh/opencybersecurityalliance/kestrel-huntbook/HEAD?filepath=tutorial) (0. Hello World Hunt). In the code block below, `<tab>` represents hitting the tab button, which calls the autocomplete function (`do_complete()`) linked above.
```
proclist = NEW process [ {"name": "cmd.exe", "pid": "123"}
                       , {"name": "explorer.exe", "pid": "99"}
                       , {"name": "firefox.exe", "pid": "201"}
                       , {"name": "chrome.exe", "pid": "205"}
                       ]
browsers = GET process FROM proclist WHERE name IN ('firefox.exe', 'chrome.exe')

# --  scenario 1
DISP <tab>                       # case 1
DISP b<tab>                      # case 2
DISP browsers<tab>               # case 3
DISP browsers <tab>              # case 4

# -- add and run this as a new block before calling DISP
abc = browsers

# --  scenario 2
DISP a<tab>                      # case 2
```

**Expected Behavior:**
For scenario 1, all cases behave as expected for autocompletion. (`suggestions` is the list returned by the `do_complete()` function)
1. `suggestions = ['TIMESTAMPED', '_', 'browsers', 'proclist']`
2. `suggestions = ['browsers']`
3. `suggestions = ['']`
[...]

PR #276, opened 2023-01-21
Autocomplete upgrade

subbyte:
Auto-complete is broken. Rewriting the auto-complete component plus adding systematic tests, which covers https://github.com/opencybersecurityalliance/kestrel-lang/issues/264

### DIFF
--- a/src/kestrel/semantics/completor.py
+++ b/src/kestrel/semantics/completor.py
@@ -214,7 +214,7 @@ def _end_with_blank_or_comma(s: str) -> bool:
 
 
 @typechecked
-def _split_last_token(s: str) -> tuple[str, str]:
+def _split_last_token(s: str) -> typing.Tuple[str, str]:
     last = ""
     if not _end_with_blank_or_comma(s):
         while not _end_with_blank_or_comma(s):
--- a/src/kestrel/semantics/completor.py
+++ b/src/kestrel/semantics/completor.py
@@ -31,6 +31,7 @@ test case in `tests/test_completion.py` to verify the logic.
 """
 
 import logging
+from typeguard import typechecked
 import re
 import lark
 import typing
@@ -53,6 +54,7 @@ _logger = logging.getLogger(__name__)
 ISO_TS_RE = re.compile(r"\d{4}(-\d{2}(-\d{2}(T\d{2}(:\d{2}(:\d{2}Z?)?)?)?)?)?")
 
 
+@typechecked
 def do_complete(
     code: str,
     cursor_pos: int,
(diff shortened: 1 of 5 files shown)


tests/test_completion.py
```diff
@@ -50,47 +129,176 @@ def a_session():
...
+    assert set(result) == set(expected)
+
+
+@pytest.mark.parametrize(
+    "code, expected",
+    [
+        ("disp ", {"conns", "_"} | TRANSFORMS),
+        ("disp c", {"onns"}),
+        ("disp conns", []),
+        ("disp conns ", {"LIMIT", "ATTR", "SORT", "WHERE", "OFFSET"}),
```
pcoccoli: Is this set `EXPRESSION_OPTIONS`?  Using the set name instead of the literals may make it easier to maintain the tests if we add more keywords to that set.

tests/test_completion.py
```diff
@@ -50,47 +129,176 @@ def a_session():
...
+        ("disp conns LIMIT 5 ", {"ATTR", "SORT", "WHERE", "OFFSET"}),
```
pcoccoli: You could use `EXPRESSION_OPTIONS - "LIMIT"` here I think (set difference)

src/kestrel/semantics/completor.py
```diff
@@ -0,0 +1,275 @@
...
+            elif token == "REVERSED":
+                expected_values.append("BY")
+            elif token == "EQUAL":
+                expected_values.append("=")
+            elif token == "TRANSFORM":
+                expected_values.extend(TRANSFORMS)
+            elif token == "ATTRIBUTE":
+                # TODO: attribute completion
+                # https://github.com/opencybersecurityalliance/kestrel-lang/issues/79
+                expected_values.append("% TODO: ATTRIBUTE COMPLETION %")
```
pcoccoli: I don't think we should actually return TODOs to the user.

pcoccoli (changes requested):

tests/test_completion.py
subbyte: Good point! Will use the `EXPRESSION_OPTIONS` for the following cases as well.

tests/test_completion.py
subbyte: Thinking the same!

src/kestrel/semantics/completor.py
subbyte: OK. I will return empty for now, and log the todo.

pcoccoli (approved):

PR #287, opened 2023-02-03
config batch size with max_result_window/pagesize and use lastsortval…

tingdai:
…ue to support pagination on large queries

### DIFF
--- a/src/kestrel_datasource_stixshifter/interface.py
+++ b/src/kestrel_datasource_stixshifter/interface.py
@@ -213,7 +213,10 @@ class StixShifterInterface(AbstractDataSourceInterface):
                     lastsortvalue = None
                     while has_remaining_results:
                         result_batch = transmission.results(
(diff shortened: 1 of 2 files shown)


src/kestrel_datasource_stixshifter/interface.py
```diff
@@ -206,19 +206,24 @@ def query(uri, pattern, session_id, config):
 
                     result_retrieval_offset = 0
                     has_remaining_results = True
+                    pagesize = transmission.get_pagesize()
```
subbyte: Shall we remove these two lines to minimize stix-shifter API change?

tingdai: Removed in bbb1bfebbb17e5e858c6b7791ea1659938fdd203

subbyte (approved):
Super clean! Cool!

PR #297, opened 2023-02-17
Initial 'fast translation' implementation in stixshifter interface

pcoccoli:


src/kestrel_datasource_stixshifter/interface.py
```diff
@@ -249,22 +255,60 @@ def query(uri, pattern, session_id, config):
...
+                    raise DataSourceError(
+                        f"STIX-shifter mapping failed with message: {mapping['error']}"
+                    )
+
+                to_stix_map = mapping["to_stix_map"]
+                df = asyncingest.translate(
+                    to_stix_map, transformers, connector_results, identity
+                )
+
+                _logger.debug("%s", df.head())
```
pcoccoli: Need to remove this

subbyte (commented):
Cool!

Two things to consider make it even better:
1. In `tests/test_fast_translate.py`, import `check_module_availability` from `kestrel_datasource_stixshifter.connector` and call `check_module_availability(connector_name)` to install the connector instead of specifying the version range in `*.yml` and `setup.cfg`. The `check_module_availability()` function will take care of version matching and install the correct version automatically. An example of a test related to it: [tests/test_stixshifter.py](https://github.com/opencybersecurityalliance/kestrel-lang/blob/develop/tests/test_stixshifter.py).
2. Add an `options` config example after [interface.py:21](https://github.com/opencybersecurityalliance/kestrel-lang/blob/develop/src/kestrel_datasource_stixshifter/interface.py#L21) so people know how to add more if needed. We can add comments to explain this is not profiles and not required. The section will be automatically compiled into our [stix-shifter interface doc](https://kestrel.readthedocs.io/en/stable/source/kestrel_datasource_stixshifter.interface.html).

pcoccoli:
We should squash this when committing; there are too many bad commits.

subbyte (approved):

PR #311, opened 2023-03-17
Support async to pipeline translation and transmission

tingdai:


subbyte:
Could you amend the commit with your github account email?

Issue #306, opened 2023-03-14
full asyncio flow to upgrade async -> sync -> async


subbyte:
We enabled firepit _fast translation_ to replace stix-shifter _result translation_ for better performance in Kestrel v1.5.10

And the enablement requires `asyncio`, so we have it in the stix-shifter datasource interface. However, the current Kestrel is not async, which means we wrap an async component into the larger sync Kestrel framework. What gets things complicated is when Kestrel is used with Jupyter, the outer layer (Jupyter) is async again.

Python asyncio is designed to have a _global event loop_, where all events are put into and scheduled by the library. However, when Kestrel executes fast translation, it needs to block it (await) until the results comes back. It cannot just let another event, e.g., Jupyter event, to take CPU, before the fast translation finishes. That means we cannot rely on the default global asyncio event scheduling, but to execute the inner async part (fast translation) into sync first, before we go to the outer async (Jupyter).

How we solve the problem is to use `nest-asyncio` for now. In the future, we should use `await` natively from `asyncio`, which requires putting the entire Kestrel runtime into async mode.

Related discussions:
- https://discuss.python.org/t/supporting-asyncio-get-event-loop-run-until-complete-in-repls/5573
- https://github.com/python/cpython/issues/93462

subbyte:
Using multiprocessing now.

PR #316, opened 2023-04-04
use produce-consume model for stixshifter transmission and translation

tingdai:


src/kestrel/datasource/manager.py
```diff
@@ -30,6 +31,9 @@ def list_data_sources_from_scheme(self, scheme):
     def query(self, uri, pattern, session_id, store):
         scheme, uri = self._parse_and_complete_uri(uri)
         i, c = self._get_interface_with_config(scheme)
-        rs = i.query(uri, pattern, session_id, c, store)
+        if scheme == "stixshifter":
```
subbyte: How about we make it generic and test any connector with [inspect.iscoroutinefunction](https://docs.python.org/3/library/inspect.html#inspect.iscoroutinefunction)?

src/kestrel/datasource/manager.py
```diff
@@ -5,6 +5,7 @@
     InvalidDataSourceInterfaceImplementation,
     ConflictingDataSourceInterfaceScheme,
 )
+import asyncio
```
subbyte, line 8: May move the `nest_asyncio` here as the first time `asyncio.run()` is called:
- https://github.com/opencybersecurityalliance/kestrel-lang/issues/306
- https://github.com/opencybersecurityalliance/kestrel-lang/blob/develop/src/kestrel_datasource_stixshifter/interface.py#L113

src/kestrel_datasource_stixshifter/interface.py
```diff
@@ -220,6 +221,25 @@ def query(uri, pattern, session_id, config, store=None):
 
             # query results should be put together; when translated to STIX, the relation between them will remain
             connector_results = []
+            queue = asyncio.Queue()
+
+            if not connector_name in config["options"]["fast_translate"]:
+                # schedule consumers
+                consumers = []
+                for _ in range(1):
```
subbyte: Any idea why we do the single cycle loop? May consider put the number of consumers as a parameter in the configuration?

src/kestrel_datasource_stixshifter/interface.py
```diff
@@ -195,7 +196,7 @@ def query(uri, pattern, session_id, config, store=None):
             check_module_availability(connector_name)
 
             data_path_striped = "".join(filter(str.isalnum, profile))
-            ingestfile = ingestdir / f"{i}_{data_path_striped}.json"
+            ingestfile = ingestdir / f"{i}_batch_index_{data_path_striped}.json"
```
subbyte: Shall we just create a function `make_ingest_stixbundle_filepath(batch_index)` instead of manipulating the string through the module? If someone forget to replace substring in the future when using `ingestfile`, that may cause issue.

subbyte (commented):
Thanks @tingdai for the first step to put Kestrel into async mode! The producer-consumer model looks good. I have some suggestions on improvements before merge.

src/kestrel_datasource_stixshifter/config.py
```diff
@@ -171,4 +175,6 @@ def load_options():
         config = {"options": {}}
     if "fast_translate" not in config["options"]:
         config["options"]["fast_translate"] = FAST_TRANSLATE_CONNECTORS
+    if "translation_consumes" not in config["options"]:
+        config["options"]["translation_consumes"] = TRANSLATION_CONSUME_NUM
```
subbyte: Suggest to put the config:
```
options:
    async_translation_workers_count: 1
```
Basically rename `consumes` to `workers` since it may be producers in the future for the second queue, the preffix `async` describes the scope, which is not multi-processing.

We may also update `TRANSLATION_CONSUME_NUM` to `ASYNC_TRANSLATION_WORKERS_CNT`.

src/kestrel/utils.py
```diff
@@ -60,6 +60,21 @@ def mkdtemp():
     return p
 
 
+def replace_path_substring(path, oldstring, newstring):
```
subbyte: May remove the function if not used.

src/kestrel/utils.py
```diff
@@ -60,6 +60,21 @@ def mkdtemp():
...
+    p = pathlib.Path(str(path).replace(oldstring, newstring))
+    return p
+
+
+def make_ingest_stixbundle_filepath(ingestfile):
```
subbyte: Could we move this function from `/src/kestrel/*` to `/src/kestrel_datasource_stixshifter/*`? These are two Python packages, which should have minimal dependencies. You may just put it in `interface.py` if the function is only used once. Or a `utils.py` module if the function is used by multiple modules in `kestrel_datasource_stixshifter`.

src/kestrel_datasource_stixshifter/interface.py
```diff
@@ -195,7 +192,9 @@ def query(uri, pattern, session_id, config, store=None):
...
+
+            ingest_stixbundle_filepath = make_ingest_stixbundle_filepath(ingestfile)
```
subbyte: Could we combine the two lines into the `make_ingest_stixbundle_filepath()` function:
```
ingestfile = ingestdir / f"{i}_batch_index_{data_path_striped}.json"
ingest_stixbundle_filepath = make_ingest_stixbundle_filepath(ingestfile)
```
and not use `replace()`?

src/kestrel_datasource_stixshifter/interface.py
```diff
@@ -243,34 +263,14 @@ def query(uri, pattern, session_id, config, store=None):
...
-                            stix_shifter_error_msg = (
-                                result_batch["error"]
-                                if "error" in result_batch
-                                else "details not avaliable"
-                            )
-                            raise DataSourceError(
-                                f"STIX-shifter transmission.results() failed with message: {stix_shifter_error_msg}"
-                            )
+                    # run the producer and wait for completion
+                    batch_index = await transmission_produce(
```
subbyte, line 285: Do we still have `connector_results` prepared for `fast_translation()`?

subbyte (changes requested):

src/kestrel_datasource_stixshifter/config.py
```diff
@@ -27,6 +28,9 @@ def set_stixshifter_logging_level():
     logging.getLogger("stix_shifter").setLevel(logging_level)
     logging.getLogger("stix_shifter_utils").setLevel(logging_level)
     logging.getLogger("stix_shifter_modules").setLevel(logging_level)
+    logging.getLogger("kestrel_datasource_stixshifter.interface").setLevel(
```
subbyte: Why do we add the `kestrel_datasource_stixshifter` package to the stix-shifter debug logging settings?

src/kestrel/datasource/manager.py
```diff
@@ -30,6 +38,9 @@ def list_data_sources_from_scheme(self, scheme):
     def query(self, uri, pattern, session_id, store):
         scheme, uri = self._parse_and_complete_uri(uri)
         i, c = self._get_interface_with_config(scheme)
-        rs = i.query(uri, pattern, session_id, c, store)
+        if inspect.iscoroutinefunction(i.query):
+            rs = asyncio.run(i.query(uri, pattern, session_id, c, store))
```
subbyte, line 42: May need to do `run_until_complete()`, need a second opinion on the logic.

src/kestrel_datasource_stixshifter/interface.py
```diff
@@ -286,41 +300,132 @@ def query(uri, pattern, session_id, config, store=None):
...
+            result_batch["data"],
+            translation_options,
+        )
+
+        if "error" in stixbundle:
+            raise DataSourceError(
+                f"STIX-shifter translation results to STIX failed with message: {stixbundle['error']}"
+            )
+
+        await asyncwrapper.SyncWrapper(store=store).cache(query_id, stixbundle)
```
subbyte, line 393: Need to check with @pcoccoli whether we have an async version of `cache()` in `firepit.aio`.

subbyte (commented):
Excellent flow now!

I have some minor comments to discuss today, plus a generic one: should we rename all `translate_consume` and similar functions to `translate_n_ingest_consume`?

src/kestrel_datasource_stixshifter/interface.py
pcoccoli: Yes, this will work.

src/kestrel_datasource_stixshifter/interface.py
```diff
@@ -310,7 +310,11 @@ async def query(uri, pattern, session_id, config, store=None):
                 consumer.cancel()
 
             # wait until all worker tasks are cancelled
-            await asyncio.gather(*consumers, return_exceptions=True)
+            responses = await asyncio.gather(*consumers, return_exceptions=True)
+            for response in responses:
+                if not isinstance(response, asyncio.CancelledError):
+                    raise response
+            print(responses)
```
pcoccoli: Please remove the print now

src/kestrel/datasource/manager.py
subbyte: Confirmed this is no longer an issue for Python>=3.7 where Kestrel currently supports. The `run()` function calls `run_until_complete()` internally: https://github.com/python/cpython/blob/3.7/Lib/asyncio/runners.py#L43

src/kestrel_datasource_stixshifter/interface.py
```diff
@@ -286,41 +300,139 @@ def query(uri, pattern, session_id, config, store=None):
...
+            # the consumers are still waiting for an item, cancel them
+            for consumer in consumers:
+                consumer.cancel()
+
+            # wait until all worker tasks are cancelled
+            responses = await asyncio.gather(*consumers, return_exceptions=True)
+            for response in responses:
+                if not isinstance(response, asyncio.CancelledError):
+                    raise response
+            print(responses)
```
subbyte: Need to be removed.

src/kestrel_datasource_stixshifter/config.py
subbyte: Actually you find a bug (not related to what you are doing) in Kestrel CLI---when `--debug` flag is used, there is no issue, but when the environment variable `KESTREL_DEBUG` is used, the debug level is not set by the [CLI](https://github.com/opencybersecurityalliance/kestrel-lang/blob/develop/src/kestrel/__main__.py).

In conclusion, it is safe to remove this line. I will fix the bug #334 separately. If you test with Kestrel CLI and `--debug` flag, it should behave as expected.

subbyte (approved):

pcoccoli (approved):

src/kestrel_datasource_stixshifter/interface.py
tingdai: It is removed in the latest commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/316/commits/817298e23031ab6a1d9cae72b5b6fc7b5eb6becf

PR #332, opened 2023-05-10
new_transformer

leila-rashidi:
I added a transformer called observed. This version requires new version of firepit. I have added a new function to firepit/sqlstorage, which is used when the new transformer is applied in Kestrel.


subbyte:
Looks like this is still not based on the latest `develop` branch, hmm...

PR #347, opened 2023-05-24
use async API from stix-shifter v5

subbyte:


src/kestrel_datasource_stixshifter/interface.py
```diff
@@ -328,14 +311,29 @@ def ingest_stixbundle_filepath(batch_index):
     return ingest_stixbundle_filepath
 
 
+async def transmission_complete(transmission, search_id):
+    status = await transmission.status_async(search_id)
+    if status["success"]:
```
pcoccoli: Why do we only check this the first time, outside of the while loop?

src/kestrel_datasource_stixshifter/interface.py
```diff
@@ -328,14 +311,29 @@ def ingest_stixbundle_filepath(batch_index):
...
+        while status["progress"] < 100 and status["status"] == "RUNNING":
+            await asyncio.sleep(1)
+            status = await transmission.status_async(search_id)
+    else:
+        stix_shifter_error_msg = (
+            status["error"] if "error" in status else "details not avaliable"
+        )
+        raise DataSourceError(
```
pcoccoli: If the second call to `status_async` fails, we won't `raise DataSourceError`.  Is that intentional?

subbyte: Good catch. Please check if the new logic is good.

pcoccoli (approved):

PR #348, opened 2023-05-30
Hotfix stixshifter 1493 aiohttp

subbyte:
This is a mitigation, not a final fix. The drawback of this hotfix is that it will let CPU-intensive translation to finish before the next round of transmission, so it nullify the advantage of the pipelining.


pcoccoli (approved):

subbyte:
This is a mitigation, not a final fix. The drawback of this hotfix is that it will let CPU-intensive translation to finish before the next round of transmission, so it nullify the advantage of the pipelining.

PR #351, opened 2023-05-31
Use background loop for query translation and transmission

cmadam:


cmadam:
This pull request will be replaced with the solution in PR https://github.com/opencybersecurityalliance/kestrel-lang/pull/355

Issue #299, opened 2023-02-20
Event as another first-class citizen


subbyte:
To simplify the design for Kestrel, we started with entity as first-class citizen. This covers horizontal lines in a [Computation Graph](https://kestrel.readthedocs.io/en/stable/theory.html) (GC). To fully express a GC, we need a way to express the vertical lines, i.e., events that connect entities.

Events are useful in connecting entities (actually entities do not connect by themselves, but always connect via events in GC). They are also useful in reasoning, e.g., viewing network connections as events between a process and an IP, the events are useful analyzing the connection frequency and pattern.

In conventional record-based reasoning, lots of data sources define each event as a record, making it a bonus: if Kestrel can express events, one can perform conventional record-based reasoning as a given. Of course, we would like to make appropriate abstraction, so the user does not need to JOIN events/records as in record-based reasoning, which requires dealing with heterogeneous types of records, which is too deep into _how to hunt_ (should be defined in compiler and auto-generated by the layer of Kestrel).

0. STIX have SCO/entity `network-traffic`, which actually is an event in its essence.
1. The support of `x-oca-event` in #295 .
2. Design the event type and make it referable (and stored in a variable) by users. More specifically, we need to upgrade `FIND` design since the most fine-grained relation will always be `entity-event-entity` as illustrated in GC.
3. Normalizing `x-oca-event`, SRO in STIX 2.1, and STIX observation (many stix-shifter connector yield an observation from a single event; STIX does not define what is the scope of observation, so this is ruled by each stix-shifter connector).

PR #354, opened 2023-06-06
Transformer ADDOBSID added

leila-rashidi:


leila-rashidi:
@subbyte  and @pcoccoli, please note that we should wait for the new version of firepit. Some of the tests are failed for this reason.

leila-rashidi:
Hi @pcoccoli, I applied comments of Dr. Shu. Can you please review the pull request?

src/kestrel/syntax/utils.py
```diff
@@ -15,7 +15,7 @@
 
 LITERALS = {"CNAME", "LETTER", "DIGIT", "WS", "INT", "WORD", "ESCAPED_STRING", "NUMBER"}
 AGG_FUNCS = {"MIN", "MAX", "AVG", "SUM", "COUNT", "NUNIQUE"}
-TRANSFORMS = {"TIMESTAMPED"}
+TRANSFORMS = {"TIMESTAMPED", "ADDOBSID"}
```
pcoccoli: This is an awkward name.

pcoccoli (approved):
I still don't like this solution.

src/kestrel/syntax/utils.py
leila-rashidi: Hi @pcoccoli and @subbyte, Do you have a better suggestion for the name of new transformer?

subbyte: `OBSERVED`?

leila-rashidi: It was the initial name, and @pcoccoli requested to change it to a better name.

pcoccoli: Since the transform only adds the ID, I guess the current name is acceptable.  I would encourage everyone to take a step back and look at what the  use case is, though.  IIRC, you're first using `TIMESTAMPED` and then this new `ADDOBSID`.  Those two perform the *exact* same operation except for the attribute name.  Couldn't you do it in a single operation by providing a list of the `observed-data` attributes you want in the result?  And if so, what you're really doing is producing a list of "records" correct?

leila-rashidi: Hi @pcoccoli, for using our kestrel analytics, we no longer need to use TIMESTAMPED transformer. However, the single operation that you suggested is great. Ideally, we can have a general transformer which takes the name of attributes that should be added as a list of attributes. If you like this solution, I can implement it in Kestrel. If we do this, in addition to the TIMSTAMPED and ADDOBSID, we have another transformer used as follows.

x=TRANSFORM(y) ATTR id, first-observed, created

Please let me know if you like this solution.

subbyte: Interesting thought on "record". To me, it is a future direction and I am viewing "record" as "event" we need to bring to Kestrel that correlate two or more entities. By defining the [event as a first-class citizen in Kestrel](https://github.com/opencybersecurityalliance/kestrel-lang/issues/299), we will be able to refer to it (such as in a variable) and pass the correlation information around.

In @leila-rashidi 's case, the data being used is already linked (as records in a single data source). However, we may think of more generic cases where users use Kestrel to correlate data from multiple data sources, then provide the linked entities to @leila-rashidi 's analytics, e.g., using `process---user-account` data from a source and `process---network-traffic` data from another source, we may output `user-account---network-traffic` data to the analytics.

subbyte:
My suggestion is to accept the current PR, while planning for the more generalized design that has `record` or `event` as first-class citizen.

For the currently PR, we need some [documentation update](https://github.com/opencybersecurityalliance/kestrel-lang/blob/develop/docs/language/eav.rst#variable-transforms) to be complete.

leila-rashidi:
Hi @subbyte, I will update documentation soon. For your information, @pcoccoli has already approved this pull request.

PR #355, opened 2023-06-08
multiprocessing stixshifter interface

subbyte:
Forget asyncio and jump to multiprocessing directly.

### Basic Design
- number of transmission processes == number of translated queries (not sure why stix-shifter translate each pattern into a list of native data source queries; so to be safe)
- number of translation processes specified by user or `cpu_count() - 2`
- serialized ingestion from translation in the main process (no need to enable multiprocess or multithreading sqlite)

### Queues
1. process queue between transmitters and translators
2. process queue between translators and ingestors

### Overhead
- lots of process spawning; looks < 1s and not a big issue
- transmission has a two level process spawning: interface process -> transmission pool process -> transmitters


subbyte:
We miss the fast translation unit test in this branch, and I don't have a good idea how to enable it. Instead, @cmadam will add test for fast translation in Kestrel integration test with Elasticsearch as live data source https://github.com/opencybersecurityalliance/federated-search-end-to-end-testing

@pcoccoli please advice if you have any good idea to do fast translation unit test here.

pcoccoli:
What is wrong with the existing fast translation test?

subbyte:
Nothing wrong, just the way the new translation function is called. I think I just come up with an idea to run it (with multiprocessing as in the interface).

subbyte:
Added 4 unit tests for the translators:
```
(stix-shifter translate, firepit fast translate) x (normal fileless execution, drop translated results into .json or .parquet)
```

The only caveat is I couldn't get `pytest-cov` working with `multiprocessing` so codecov does not know that the `translator.py` file is 100% covered.

pcoccoli (approved):

PR #365, opened 2023-06-16
Limit in get find

cmadam:


src/kestrel_datasource_stixshifter/query.py
```diff
@@ -113,10 +121,15 @@ def query_datasource(uri, pattern, session_id, config, store):
                 config["options"]["translation_workers_count"],
                 dsl["queries"],
                 raw_records_queue,
+                profile_limit,
             ):
                 for _ in range(config["options"]["translation_workers_count"]):
                     for packet in iter(translated_data_queue.get, STOP_SIGN):
                         if packet.success:
+                            num_objects = len(packet.data.get("objects", []))
```
subbyte: `packet` can either be STIX bundle or DataFrame, depending on the option in translation. We may need to handle both cases. Check the `ingest()` function to better see the formats.

I suggest to put this section into a function like `ingest()` when it gets larger with two options.

src/kestrel/codegen/commands.py
```diff
@@ -462,9 +466,7 @@ def find(stmt, session):
                 _logger.debug(
                     f"merge {local_var_table} and {prefetch_ret_entity_table} into {return_var_table}."
                 )
-                session.store.merge(
-                    return_var_table, [local_var_table, prefetch_ret_entity_table]
-                )
+                session.store.merge(return_var_table, [prefetch_ret_entity_table])
```
subbyte: Is it possible to just abandon `merge()`? Let's discuss in today's call.

src/kestrel_datasource_stixshifter/worker/transmitter.py
```diff
@@ -159,12 +167,20 @@ def retrieve_data(self):
...
+                            batch_size = self.limit - result_retrieval_offset
+                            if batch_size > self.retrieval_batch_size:
+                                batch_size = self.retrieval_batch_size
                 else:
                     has_remaining_results = False
 
                 is_retry_cycle = False
 
             else:
+                has_remaining_results = False
```
subbyte: Thanks for catching this logic bomb!

However, this statement should be put in the second subcase to enable retry (the first subcase) as added in the latest develop: https://github.com/opencybersecurityalliance/kestrel-lang/blob/develop/src/kestrel_datasource_stixshifter/worker/transmitter.py#L210

subbyte (changes requested):

src/kestrel_datasource_stixshifter/worker/transmitter.py
cmadam: Fixed in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/365/commits/e58be87e7557a852f09d08a294f68b711fd91c19

src/kestrel/codegen/commands.py
cmadam: Fixed in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/365/commits/17b344335a9db2552db59a9c541577ee9f2a56de

src/kestrel_datasource_stixshifter/query.py
cmadam: Fixed in https://github.com/opencybersecurityalliance/kestrel-lang/pull/365/commits/6be265cf4490884e7dcc703ac2aee4a29a936db1

subbyte (approved):

Issue #371, opened 2023-07-10
Update the terminals in grammar file with lower cases to workaround the bug of `lark-js`


jillyj:
**Is your feature request related to a problem? Please describe.**
When generating the JS parser for Kestrel grammar, I encountered an issue while using the generated parser. It throws "invalid regular expression" error. Please refer to the bug I opened in `lark-js` repo. https://github.com/lark-parser/Lark.js/issues/21

**Describe the solution you'd like**
The workaround is to change some terminals from upper case to lower case which makes the parser working.

**Describe alternatives you've considered**

**Additional context**

jillyj:
Here is the updated grammar that `lark-js` could generate correct parser.
```
//
// Kestrel Grammar (version 1.7.0 - 2023/06/15)
//

//
// A huntflow is a sequence of statements
//

start: statement*

statement: assignment
         | command_no_result

// If no VARIABLE is given, default to _ in post-parsing
// For assign or merge, the result variable is required
// This eliminates meaningless huntflows like `var1 var2 var3`
assignment: VARIABLE "=" expression
          | VARIABLE "=" VARIABLE ("+" VARIABLE)+
          | (VARIABLE "=")? command_with_result

// "?" at the beginning will inline command
?command_with_result: find
                    | get
                    | group
                    | join
                    | load
                    | new
                    | sort

?command_no_result: apply
                  | disp
                  | info
                  | save

//
// All commands
//

find: "FIND"i ENTITY_TYPE RELATION (REVERSED)? VARIABLE where_clause? timespan?

get: "GET"i ENTITY_TYPE ("FROM"i datasource)? where_clause timespan?

group: "GROUP"i VARIABLE BY grp_spec ("WITH"i agg_list)?

join: "JOIN"i VARIABLE "," VARIABLE (BY ATTRIBUTE "," ATTRIBUTE)?

load: "LOAD"i stdpath ("AS"i ENTITY_TYPE)?

new: "NEW"i ENTITY_TYPE? var_data

sort: "SORT"i VARIABLE BY ATTRIBUTE (ASC|DESC)?

apply: "APPLY"i analytics_uri "ON"i variables ("WITH"i args)?

disp: "DISP"i expression

info: "INFO"i VARIABLE

save: "SAVE"i VARIABLE "TO"i stdpath

//
// Variable definition
//

variables: VARIABLE ("," VARIABLE)*

VARIABLE: CNAME

//
// Expression
//

[...]

PR #372, opened 2023-07-11
kestrel grammar: change some terminals from upper case to lower case

jillyj:
To fix issue #371.

### DIFF
--- a/src/kestrel/syntax/kestrel.lark
+++ b/src/kestrel/syntax/kestrel.lark
@@ -10,7 +10,7 @@ start: statement*
 
 statement: assignment
          | command_no_result
-         
+
 // If no VARIABLE is given, default to _ in post-parsing
 // For assign or merge, the result variable is required
 // This eliminates meaningless huntflows like `var1 var2 var3`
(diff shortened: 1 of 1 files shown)


subbyte:
Could you add some corresponding functions in `parser.py` to handle the new rules (converted from terminals) @jillyj ?

jillyj:
> Could you add some corresponding functions in `parser.py` to handle the new rules (converted from terminals) @jillyj ?

@pcoccoli could you please help to do this?

pcoccoli:
I can look at it next week.

jillyj:
@pcoccoli Thank you very much~

src/kestrel/syntax/kestrel.lark
```diff
@@ -113,23 +113,23 @@ comparison_null: ENTITY_ATTRIBUTE_PATH NULL_OP NULL
...
-         | "t\"" ISOTIMESTAMP "\""
-         | "t'"  ISOTIMESTAMP "'"
+timestamp:       isotimestamp
+         | "\""  isotimestamp "\""
+         | "'"   isotimestamp "'"
+         | "t\"" isotimestamp "\""
+         | "t'"  isotimestamp "'"
 
-ISOTIMESTAMP: /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d(\.\d+)?Z/
+isotimestamp: /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d(\.\d+)?Z/
```
pcoccoli, line 132: Are you sure this one needs to be a rule?  It doesn't use alternation at all; seems like it could stay as a token (in all caps).

pcoccoli:
Obsoleted by #378

PR #377, opened 2023-07-17
fix #254 with a command decorator

### LINKED ISSUE #254: Allow empty input variables
Currently Kestrel will raise an exception when the "input" variable for a statement has no entities (i.e. is empty).  That stops the hunt with an error, and makes it look like something has failed when really it may just mean that the hunt hasn't found what it's looking for.  

It may be better to just accept empty vars, and let the user decide what to make of it.

**Describe alternatives you've considered**
Maybe a "warning" message?  Probably not helpful.

**Additional context**
Example:

```
procs = GET process FROM file:///home/pcoccoli/tmp/bundle.json
             WHERE [process:command_line LIKE 'wont_match_anything']
proc_conns = FIND network-traffic CREATED BY procs
```

Results in:
```
13:12:38 DEBUG kestrel.codegen.commands Executing 'find' with statement: {'command': 'find', 'type': 'network-traffic', 'relation': 'created', 'reversed': True, 'input': 'procs', 'timerange': None, 'output': 'proc_conns'}
  [omitted for brevity]
  File "/home/pcoccoli/github/kestrel-lang/src/kestrel/codegen/commands.py", line 84, in wrapper
    raise EmptyInputVariable(varname)
kestrel.exceptions.EmptyInputVariable: [ERROR] EmptyInputVariable: empty input variable "procs"
rewrite the command to use a non-empty variable as input.
```
The `find` command is the one raising the exception.

Since the exception stops the hunt early, it's harder to build automation around a hunt.

subbyte:


pcoccoli (approved):
We may need some additional tests with empty input, e.g. for `GROUP`.  I'm not sure what will happen.

subbyte:
The tests for GET/FIND/DISP cover three types of commands (with/without input/output) for `@_skip_command_if_empty_input` to be complete.

Just added some more tests for GROUP/ASSIGN for double check.

PR #378, opened 2023-07-18
Change some terminals to rules in the grammar

### LINKED ISSUE #371: Update the terminals in grammar file with lower cases to workaround the bug of `lark-js`

pcoccoli:
Fixes #371

### DIFF
--- a/src/kestrel/syntax/kestrel.lark
+++ b/src/kestrel/syntax/kestrel.lark
@@ -10,7 +10,7 @@ start: statement*
 
 statement: assignment
          | command_no_result
-         
+
 // If no VARIABLE is given, default to _ in post-parsing
 // For assign or merge, the result variable is required
 // This eliminates meaningless huntflows like `var1 var2 var3`
@@ -46,7 +46,7 @@ join: "JOIN"i VARIABLE "," VARIABLE (BY ATTRIBUTE "," ATTRIBUTE)?
 
 load: "LOAD"i stdpath ("AS"i ENTITY_TYPE)?
 
-new: "NEW"i ENTITY_TYPE? VAR_DATA
+new: "NEW"i ENTITY_TYPE? var_data
 
 sort: "SORT"i VARIABLE BY ATTRIBUTE (ASC|DESC)?
 
@@ -104,8 +104,8 @@ offset_clause: "OFFSET"i INT
            | comparison_null
            | "(" disjunction ")"
 
-comparison_std:  ENTITY_ATTRIBUTE_PATH OP      value
-comparison_null: ENTITY_ATTRIBUTE_PATH NULL_OP NULL
+comparison_std:  ENTITY_ATTRIBUTE_PATH op      value
+comparison_null: ENTITY_ATTRIBUTE_PATH null_op NULL
 
 //
 // Timespan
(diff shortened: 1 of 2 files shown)


src/kestrel/syntax/kestrel.lark
```diff
@@ -228,24 +226,18 @@ NOT: "NOT"i
...
+  | NOT? op_keyword
 
-OP_SIGN: "="
-       | "=="
-       | "!="
-       | ">"
-       | "<"
-       | ">="
-       | ">="
+OP_SIGN: /([!=]?=|[<>]=?)/
```
subbyte, line 232: rule or terminal? I think you add a new rule in `parser.py`

src/kestrel/syntax/kestrel.lark
```diff
@@ -228,24 +226,18 @@ NOT: "NOT"i
...
 
-OP_KEYWORD: IN
+op_keyword: IN
           | LIKE
           | MATCHES
           | ISSUBSET
           | ISSUPERSET
 
-NULL_OP: IS (WS+ NOT)?
+null_op: IS (NOT)?
```
subbyte: Could we remove the parenthesis to `IS NOT?` ?

src/kestrel/syntax/parser.py
```diff
@@ -398,6 +445,18 @@ def args(self, args):
     def arg_kv_pair(self, args):
         return {_first(args): args[1]}
 
+    def op(self, args):
+        return " ".join([arg.upper() for arg in args])
+
+    def op_sign(self, args):
+        return _first(args) if args else ""
```
subbyte: Why do we need `if args else ""`? Could an `op_sign` be empty?

src/kestrel/syntax/kestrel.lark
```diff
@@ -228,24 +226,18 @@ NOT: "NOT"i
 ISSUBSET: "ISSUBSET"i
 ISSUPERSET: "ISSUPERSET"i
 
-OP: OP_SIGN
-  | (NOT WS+)? OP_KEYWORD
+op: OP_SIGN
```
subbyte, line 229: Associated with below (rule or terminal)

src/kestrel/syntax/parser.py
```diff
@@ -328,8 +335,48 @@ def timespan_absolute(self, args):
...
+
+    def json_pair(self, args):
+        v = _first(args)
+        if "ESCAPED_STRING" in args[0].type:
+            v = unescape_quoted_string(v)
+        return v, args[1]
+
+    def json_value(self, args):
+        v = _first(args)
+        if "ESCAPED_STRING" in args[0].type:
```
subbyte: Could we use the explicit comparison as `NUMBER` below, i.e., `args[0].type == self.token_prefix + "ESCAPED_STRING"` ?

subbyte (commented):
Thanks for the fix! I go through the change and may find some typos and simplifications.

src/kestrel/syntax/kestrel.lark
pcoccoli: I believe it can stay a terminal, so I'll remove the rule method in the Transformer in parser.py.

src/kestrel/syntax/kestrel.lark
pcoccoli: I think so, yes.

src/kestrel/syntax/parser.py
pcoccoli: `op_sign` will be removed as it's not necessary.

src/kestrel/syntax/kestrel.lark
pcoccoli: `OP_SIGN` will remain a terminal.

src/kestrel/syntax/parser.py
pcoccoli: I this in order to match `ESCAPED_STRING` or `ESCAPED_STRING_WS`, but the `_WS` version shouldn't ever show up here.  I will change it.

subbyte (approved):
Looks good!

PR #392, opened 2023-08-01
add custom pattern and stop now to ss-diag

subbyte:


tests/utils.py
```diff
@@ -48,3 +48,26 @@ def stixshifter_profile_lab101(tmp_path):
...
+        connector: elastic_ecs
+        connection:
+            host: elastic.securitylog.company.com
+            port: 9200
+            selfSignedCert: false
+            indices: host101
+        config:
+            auth:
+                id: VuaCfGcBCdbkQm-e5aOx
+                api_key: ui2lp2axTNmsyakw9tvNnw
```
pcoccoli, line 70: I assume these are not real credentials?

pcoccoli (approved):

tests/utils.py
subbyte: Right. Same as in our config doc.

PR #395, opened 2023-08-16
KaaS Image Build and Push to Dockerhub

kpeeples:


.github/workflows/kaas-docker-image.yml
```diff
@@ -0,0 +1,34 @@
+name: Docker image
+on:
+  push:
```
subbyte: How about we follow the same rules as in https://github.com/opencybersecurityalliance/kestrel-lang/blob/develop/.github/workflows/publish-to-pypi.yml#L4 to trigger this---release a new container image when we do a new kestrel-lang release?

src/containers/docker/README.rst
subbyte, line 1: Could you create a new directory `/docs/deployment/` and move this rst file to there? Then you can add `index.rst` to the new directory and a pointer in https://github.com/opencybersecurityalliance/kestrel-lang/blob/develop/docs/index.rst to list this doc under menu "deployment".

Once this PR is merged, readthedocs will automatically trigger a build of the latest doc https://kestrel.readthedocs.io/en/latest/ (for the `develop` branch of `kestrel-lang`) to include your doc in official Kestrel doc site. The stable version of the doc will be updated when we make the next Kestrel release.

src/containers/oci/README.rst
subbyte, line 1: I believe this can also go to the `/docs/deployment/` (maybe in `/docs/deployment/index.rst`) as a comment after the content listed.

subbyte (commented):
Thanks for the Dockerhub PR @kpeeples ! This is great, and I have a few comments to make it better organized.

.github/workflows/kaas-docker-image.yml
kpeeples: ok done

src/containers/docker/README.rst
kpeeples: done

src/containers/oci/README.rst
kpeeples: done

subbyte (approved):

PR #403, opened 2023-09-17
Add rudimentary interactive CLI ikestrel

pcoccoli:
Includes command line history and editing, as well as tab completion.

### DIFF
--- a/bin/ikestrel
+++ b/bin/ikestrel
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+
+# Executing a hunt in an interactive CLI
+# Usage: `ikestrel [-v] [--debug]`
+
+import argparse
+import cmd
+import logging
+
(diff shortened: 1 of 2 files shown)


bin/ikestrel
```diff
@@ -0,0 +1,92 @@
...
+        super().__init__()
+
+    def default(self, line: str):
+        try:
+            outputs = self.session.execute(line)
+            display_outputs(outputs)
+        except KestrelException as e:
+            print(e)
+
+    def completenames(self, text, *ignored):
```
subbyte, line 52: Is this a virtual method to override? I could not find it at https://docs.python.org/3/library/cmd.html

subbyte (commented):
Good idea! Just a question.

subbyte (approved):

PR #414, opened 2023-11-06
Add STIX-to-OCSF entity and property mapping

cmadam:


packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/property/stix.yaml
```diff
@@ -0,0 +1,196 @@
...
+email-message:to_refs[*].value: email.to
+email-message:cc_refs[*].value: email.cc
+email-message:subject: email.subject
+# email-message:received_lines
+email-message:additional_header_fields: email.raw_header
+# email-message:body
+# email-message:body_multipart,
+# email-message:raw_email_ref
+
+# email-mime-part-type:body: file.mime_type
```
subbyte: Is this `email-message:body_multipart.body`?

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/property/stix.yaml
```diff
@@ -0,0 +1,196 @@
...
+# email-mime-part-type:body_raw_ref
```
subbyte: We need to add this `email-message:body_multipart.body_raw_ref.name: file.name` for event [Email File Activity](https://schema.ocsf.io/1.0.0/classes/email_file_activity?extensions=)

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/property/stix.yaml
```diff
@@ -0,0 +1,196 @@
...
+# email-mime-part-type:content_type
+# email-mime-part-type:content_disposition
+
+file:accessed: file.accessed_time
+file:created: file.created_time
+file:name: file.name
+file:size: file.size
+file:hashes.SHA-256: file.hashes.SHA-256
```
subbyte: This may be an example that we need more complicated mapping. Does OCSF allow `file.hashes.SHA-256`? or it needs to be `file.hashes = [{"algorithm_id": 3, "value": "xxx"}]`.

If we use JMESPath, the query should be `file.hashes[?algorithm_id == 3].value`.

This is an important item to discuss: if we need data mapping as well (STIX bundle to OCSF), it may be more complicated than field/property mapping, and it may be different from query mapping.

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/property/stix.yaml
```diff
@@ -0,0 +1,196 @@
...
+# network_traffic:ipfix
+# network_traffic:src_payload_ref
+# network_traffic:dst_payload_ref
+# network_traffic:encapsulates_refs
+# network_traffic:encapsulated_by_ref
+
+process:binary_ref.name: file.name
+process:command_line: 
+  - process.cmd_line
+  - actor.process.cmd_line
```
subbyte: This is tricky. Depending on whether we are querying `Network Activity` or `Process Activity`, `actor.process` could mean different things and it should or should not be included. Discussion needed.

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/entity/stix.yaml
```diff
@@ -0,0 +1,24 @@
...
+network-traffic: network-activity
+process: process
+software: product
+url: http_request
+user-account: user
+x-ibm-finding: security-finding
+x-ibm-ttp-tagging: attack
+x-oca-asset: device
+# TODO: mapping for x-oca-event
+x-oca-event:
```
subbyte: How about mapping it to [Base Event](https://schema.ocsf.io/1.0.0/base_event?extensions=)?

I am thinking to add two alias to it in Kestrel: `event` and `activity`.

The only issue is this is good for query translation, not data translation (another example below on `file.hashes`).

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/entity/stix.yaml
```diff
@@ -0,0 +1,24 @@
...
+autonomous-system: organization
+directory: file
+domain-name: endpoint
+email-addr: user
+email-message: email
+file: file
+ipv4-addr: endpoint
+ipv6-addr: endpoint
+mac-addr: endpoint
+network-traffic: network-activity
```
subbyte: I am rethinking about whether we should use dash or underscore. Looks like the URLs of all activities uses underscore, and we may better do the underscore.

We can also auto-normalize dash to underscore in Kestrel parser.

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/property/stix.yaml
```diff
@@ -0,0 +1,196 @@
+autonomous-system:name: organization.name
+autonomous-system:number: organization.uid
+
+directory:path: file.path
+directory:accessed: file.accessed_time
+directory:created: file.created_time
+directory:modified: file.modified_time
+
+domain-name.value:
```
subbyte, line 12: Potential an issue for data translation. See `file.hashes` for detailed example.

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/property/stix.yaml
```diff
@@ -0,0 +1,196 @@
...
+x-oca-asset:os_name: device.os.name
+x-oca-asset:hostname: device.hostname
+x-oca-asset:device_id: device.uid
+x-oca-asset:ip_refs[*].value: device.network_interfaces[*].ip
+x-oca-asset:mac_refs[*].value: device.network_interfaces[*].mac
+x-oca-asset:os_ref: device.os
+x-oca-asset:architecture: device.hw_info
+x-oca-asset:host_type: device.type
+x-oca-asset:ingress: device.network_interfaces
+x-oca-asset:egress:	device.network_interfaces
```
subbyte: typo fix: tab to space

subbyte (commented):
Thank you for adding the long list of mapping, @cmadam !

I go through them and add comments. The most important thing I realize is we may need more complicated structure for data mapping (in here or separate yaml). The good news is while we don't know the data (or partial data) in query mapping (so we could not do condition), we may be able to do condition-based mapping for data mapping.

By data mapping, I am referring to STIX bundle to OCSF mapping used by the `cache` module. We may defer it to the stix-shifter data source interface to do the data mapping for STIX bundle. Then we need two mappings from each interface:
- query mapping from OCSF to the target language, e.g., SQL/KQL/STIX.
- data mapping from the source data, e.g., STIX bundle, to OCSF
I didn't expect, but this structure of mapping share high-level common ideas with stix-shifter. I think we should get Paul's opinion on the tabular format of STIX bundle, which could simplify the mapping from STIX SCOs to OCSF activities/records.

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/property/stix.yaml
subbyte: We decided to only execute process query on _Process Activity_ events and only execute network-traffic query on _Network Activity_ events (category 4 with multiple event types in OCSF). So just remove `actor.process.cmd_line` to avoid different semantics of `actor` in the two types of events.

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/property/stix.yaml
cmadam: Added in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/414/commits/15c9caec06163a26299565dfdf151fcf4509fcfa

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/property/stix.yaml
cmadam: Added hashes mapping using JMESPath in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/414/commits/15c9caec06163a26299565dfdf151fcf4509fcfa

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/property/stix.yaml
cmadam: Removed `actor.process.cmd_line` in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/414/commits/15c9caec06163a26299565dfdf151fcf4509fcfa

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/entity/stix.yaml
cmadam: Mapped `x-oca-event` to `event` and `activity` (virtual) classes in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/414/commits/15c9caec06163a26299565dfdf151fcf4509fcfa

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/entity/stix.yaml
cmadam: Replaced `-` with `_` in several OCSF entities, including `network_activity` in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/414/commits/15c9caec06163a26299565dfdf151fcf4509fcfa

packages-nextgen/kestrel_core/src/kestrel/frontend/mapping/property/stix.yaml
cmadam: Fixed typo in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/414/commits/15c9caec06163a26299565dfdf151fcf4509fcfa

subbyte (commented):
This is now a comprehensive mapping with mappings and metadata (which activity to map to). Looks very good.

Last item to update: In the `property/stix.yaml` mapping, there are comments at some sections not referring to an OCSF activity, but an OCSF object. We should identify the activities for them and replace the comment `Device Object [20]` with activities where the object exist (and the mapping may need a prefix like `device.` depending on the activity).

subbyte (approved):

PR #417, opened 2023-11-13
Initial Kestrel2 filter IR code

pcoccoli:


packages-nextgen/kestrel_core/src/kestrel/ir/instructions.py
```diff
@@ -0,0 +1,20 @@
...
+from kestrel.ir.filter import (
+    IntComparison,
+    FloatComparison,
+    StrComparison,
+    ListComparison,
+    BoolExp,
+)
+
+
+class IRNode(DataClassJSONMixin):
```
subbyte: Two thoughts:
1. Do you prefer the base class name `IRNode` or `Instruction`? I am good with either.
2. Could we add a property field `id()` to the base class which can simply be `return hash(self)`? This is required by the `IRGraph` class I am writing. I will add more requirements later.

subbyte (commented):
Looks good, just two comments on `IRNode` before merge.

packages-nextgen/kestrel_core/src/kestrel/ir/instructions.py
pcoccoli: 1. I suppose `Instruction` is maybe nicer, and less redundant since the module name is already `ir`.
2. Let's handle the `id()` in a different PR.

subbyte: Sounds good.

subbyte (approved):

PR #419, opened 2023-11-16
init IRGraph and Instruction classes

subbyte:


packages-nextgen/kestrel_core/src/kestrel/exceptions.py
```diff
@@ -0,0 +1,29 @@
...
+class InvalidInstruction(KestrelError):
+    pass
+
+class InvalidSeralizedGraph(KestrelError):
+    pass
+
+class InvalidSeralizedInstruction(KestrelError):
+    pass
+
+class VariableNotExist(KestrelError):
```
pcoccoli: I know we have this exception already in Kestrel, but I find the name doesn't read well.  "Not exist" is odd phrasing; maybe `NonexistentVariable` or `VariableNotFound`?

packages-nextgen/kestrel_core/src/kestrel/ir/graph.py
```diff
@@ -0,0 +1,351 @@
...
+        return [
+            n
+            for n in nodes
+            if all([getattr(n, k, None) == v for (k, v) in attr2val.items()])
+        ]
+
+    def get_variables(self) -> Iterable[Variable]:
+        """Get all variables
+
+        This method returns a list variables, equivalent to *Symbol Table* used in traditional (non-graph-IR) language compilers.
```
pcoccoli: Should be "list of variables"

pcoccoli (commented):
I suggest adding some unit tests.

packages-nextgen/kestrel_core/src/kestrel/ir/instructions.py
```diff
@@ -10,11 +20,66 @@
...
+@dataclass
+class Variable(Instruction):
+    name: str
+    deceased: bool = False
+
+
+@dataclass
+class Source(Instruction):
+    uri: str
+    interface: str
```
pcoccoli, line 59: Is `interface` redundant since we have `uri` already?  Seems like we would either store `interface` and `name` or just `uri` (which is of the form "<interface>://<name>").

subbyte: I try to solve it with the best solution I think: using `InitVar` in `dataclasses` for `uri`, so we can use `uri` in the constructor. However, I encountered an issue when combining `InitVar` with `mashumaro` (serialization or deserialization): https://github.com/Fatal1ty/mashumaro/issues/179

So I just create `source_from_uri()` as explicit constructor and keep the `Source` class simple for now.

PR #421, opened 2023-11-20
K2 initial parser rebased 2

pcoccoli:


cmadam (commented):
The changes in this commit make sense to me.  First, the code re-org  breaks the  `parser.py` file into `parser.py` and `compile.py` files and it also  moves the `load_data_file` function into the `packages-nextgen/kestrel_core/src/kestrel/utils.py` file.  Second, the code creates one single parser (as opposed to creating a new parser that loads the grammar for every run).  Third, the test file contains a minimal unit test, but as we fix more of the code, the other tests will be gradually re-introduced.

pcoccoli:
I want to improve this PR by integrating with Xiaokui's IRGraph work.  I should have an update shortly.

subbyte (approved):
Looks good. I will add/revise some graph building logic after this is merged.

PR #428, opened 2023-11-29
kestrel2: Add MultiComp

pcoccoli:


cmadam (commented):
This makes sense to me.  We will have to import the new `MultiComp` class in `compile.py`.  This simplifies the representation of the filter. @subbyte do you see any undesirable effects from using `MultiComp` in the structure of the `Filter.exp` field?

PR #432, opened 2023-11-29
kestrel2 Limit instruction

pcoccoli:


packages-nextgen/kestrel_core/src/kestrel/frontend/compile.py
```diff
@@ -90,9 +91,14 @@ def get(self, args):
...
         projection_node = graph.add_node(ProjectEntity(args[0].value), filter_node)
+        root = projection_node
         if len(args) > 3:
-            filter_node.timerange = args[3]
-        return graph, projection_node
+            for arg in args[3:]:
+                if isinstance(arg, TimeRange):
+                    filter_node.timerange = args[3]
+                elif isinstance(arg, Limit):
+                    root = graph.add_node(arg, projection_node)
```
subbyte, line 100: The current order is `Source -> Filter -> Projection -> Limit/TimeRange`.

Shall we do `Source -> Limit/TimeRange -> Filter -> Projection`?

subbyte (approved):

PR #434, opened 2023-11-30
Mapping for entity names and properties

cmadam:


packages-nextgen/kestrel_core/src/kestrel/frontend/compile.py
```diff
@@ -88,8 +133,13 @@ def assignment(self, args):
 
     def get(self, args):
         graph = IRGraph()
+        entity_name = args[0].value
+        mapped_entity_name = self.entity_map.get(entity_name, entity_name)
+        filter = args[2]
```
pcoccoli: Might want to avoid the variable name `filter` since it's a global function in Python

packages-nextgen/kestrel_core/src/kestrel/frontend/compile.py
```diff
@@ -61,17 +62,61 @@ def _create_comp(field: str, op: str, value):
...
 
 
+def _map_filter_exp(entity_name, filter_exp, property_map):
+    if isinstance(filter_exp, (IntComparison, FloatComparison, StrComparison,
+                               ListComparison)):
+        # get the field
+        field = filter_exp.field
+        # add entity to field if it doesn't have one already
+        if ':' not in field:
+            field = f"{entity_name}:{field}"
```
pcoccoli, line 91: What will happen here if the user *isn't* using STIX objects and properties?  I still don't know OCSF that well, but if the statement is using OCSF it seems like we'll be creating an invalid field here.
Maybe if the new field name we created isn't in the map, we should use the original field (instead of the one with `{entity_name}:` prepended?

packages-nextgen/kestrel_core/src/kestrel/frontend/compile.py
cmadam: Addressed in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/434/commits/1a375dc2d5bb98b706c806624e1552c20a660b12

packages-nextgen/kestrel_core/src/kestrel/frontend/compile.py
cmadam: > Maybe if the new field name we created isn't in the map, we should use the original field (instead of the one with {entity_name}: prepended?

I addressed this issue in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/434/commits/1a375dc2d5bb98b706c806624e1552c20a660b12

To fully address this comment, we need to provide a way, in  the Kestrel  configuration, to provide front-end mapping between multiple standards, in addition to the existing STIX to OCSF mapping.

pcoccoli (approved):
looks good

packages-nextgen/kestrel_core/tests/test_parser.py
```diff
@@ -73,3 +74,51 @@ def test_parser_get_with_limit(stmt, expected):
     assert len(limits) == 1
     limit = limits[0]
     assert limit.num == expected
+
+
+def get_parsed_filter_exp(stmt):
+    parse_tree = parse_kestrel(stmt)
+    parse_tree_string = parse_tree.to_json()
+    parse_tree_json = json.loads(parse_tree_string)
```
subbyte: Can we try `to_dict()` directly instead of `to_json()` and `loads()` here?

https://github.com/opencybersecurityalliance/kestrel-lang/blob/develop/packages-nextgen/kestrel_core/src/kestrel/ir/graph.py#L477

A even simpler way is use the IRGraph directly instead of dict. You may try
```
filter_node = parse_tree.get_nodes_by_type(Filter).pop()
filter_exp = filter_node.exp
```
Then `parse_filter = get_parsed_filter_exp(stmt)` will get you a `kestrel.ir.filter.*` type to verify directly (not a dict). You can do `to_dict()` for the object if needed.

packages-nextgen/kestrel_core/src/kestrel/frontend/compile.py
```diff
@@ -61,17 +62,61 @@ def _create_comp(field: str, op: str, value):
     return comp
 
 
+def _map_filter_exp(entity_name, filter_exp, property_map):
```
subbyte: Can we add types for the arguments and return, and the `@typechecked` decorator?

packages-nextgen/kestrel_core/src/kestrel/frontend/parser.py
```diff
@@ -1,16 +1,42 @@
...
 from kestrel.frontend.compile import _KestrelT
 from kestrel.utils import load_data_file
+from lark import Lark
+import os
+import yaml
+
+frontend_mapping = {}
+
+
+def get_mapping(mapping_type, mapping_package, mapping_filepath):
```
subbyte: Can we add typing for the arguments/return?

subbyte (commented):
Thank you @cmadam ! The logic is good, and I added some comments around typing and simplification.

packages-nextgen/kestrel_core/tests/test_parser.py
cmadam: Addressed in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/434/commits/182d42e746fc877f7fc111550e699d126b17aa00

packages-nextgen/kestrel_core/src/kestrel/frontend/compile.py
cmadam: Addressed in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/434/commits/182d42e746fc877f7fc111550e699d126b17aa00

packages-nextgen/kestrel_core/src/kestrel/frontend/parser.py
cmadam: Addressed in commit https://github.com/opencybersecurityalliance/kestrel-lang/pull/434/commits/182d42e746fc877f7fc111550e699d126b17aa00

subbyte (approved):
Awesome!

PR #436, opened 2023-12-01
Preserve sort order for indirect attributes

### LINKED ISSUE #435: Sorting by indirect attributes doesn't work
**Describe the bug**
Sort order is lost in DISP for indirect attributes

**Details of the bug**
```
p = GET process
    FROM file:///home/pcoccoli/github/kestrel-lang/test-data/doctored-1k.json
    WHERE name in ('conhost.exe', 'services.exe', 'taskhost.exe')
DISP p ATTR binary_ref.name SORT BY binary_ref.name DESC
sp = SORT p BY binary_ref.name DESC
DISP sp ATTR binary_ref.name
```
The two DISP outputs do not match - the first is sorted correctly; the second is not.

**To Reproduce**
Run the statements above

**Expected behavior**
The two DISP outputs should be the same - sorted descending.

**Screenshots**
```
binary_ref.name
   taskhost.exe
   services.exe
    conhost.exe

binary_ref.name
   services.exe
    conhost.exe
   taskhost.exe

[SUMMARY] block executed in 1 seconds
VARIABLE    TYPE  #(ENTITIES)  #(RECORDS)  directory*  file*  process*  user-account*
       p process            9           9           8     16        23              8
      sp process            9           9           0      0         0              0
*Number of related records cached.
```

pcoccoli:
Fixes #435

### DIFF
--- a/packages/kestrel_core/tests/test_command_assign.py
+++ b/packages/kestrel_core/tests/test_command_assign.py
@@ -99,3 +99,10 @@ def test_assign_from_empty_var(proc_bundle_file):
         q = s.symtable["q"]
         assert len(q) == 0
         assert q.records_count == 0
+
+
+def test_assign_sort_by_indirect_attr(proc_bundle_file):
+    with Session() as s:
+        s.execute(f"c = GET network-traffic FROM file://{proc_bundle_file} WHERE src_port > 0")
+        s.execute("d = c SORT BY dst_ref.value")
+        _ = s.get_variable("d")
--- a/packages/kestrel_core/pyproject.toml
+++ b/packages/kestrel_core/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
     "pandas>=2.0.3",
     "pyarrow>=13.0.0",
     "tabulate>=0.9.0",
-    "firepit>=2.3.29",
+    "firepit>=2.3.30",
 ]
 
 [project.optional-dependencies]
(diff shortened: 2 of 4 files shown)


packages/kestrel_core/tests/test_command_assign.py
```diff
@@ -99,3 +99,10 @@ def test_assign_from_empty_var(proc_bundle_file):
         q = s.symtable["q"]
         assert len(q) == 0
         assert q.records_count == 0
+
+
+def test_assign_sort_by_indirect_attr(proc_bundle_file):
+    with Session() as s:
+        s.execute(f"c = GET network-traffic FROM file://{proc_bundle_file} WHERE src_port > 0")
+        s.execute("d = c SORT BY dst_ref.value")
+        _ = s.get_variable("d")
```
subbyte, line 108: Do we need a `DISP d ATTR dst_ref.value` to check?

subbyte (approved):
May add an additional assert in test.

packages/kestrel_core/tests/test_command_assign.py
pcoccoli: That's covered in the test in test_command_disp.py.  `s.get_variable("d")` was raising an `InvalidAttr` exception, so this test just covers that code path.

subbyte: I see. Then I have no concern.

PR #444, opened 2023-12-11
Initial sql support for kestrel2

pcoccoli:
Implements a basic Kestrel 2 instruction to SQL translator.
Implements a sqlite3-based cache.

### DIFF
--- a/packages-nextgen/kestrel_core/src/kestrel/cache/sqlite.py
+++ b/packages-nextgen/kestrel_core/src/kestrel/cache/sqlite.py
@@ -13,12 +13,11 @@ from kestrel.interface.datasource.codegen.sql import SqlTranslator
 from kestrel.ir.graph import IRGraphSoleInterface
 from kestrel.ir.instructions import (
     Construct,
-    Filter,
     Instruction,
     Return,
     SourceInstruction,
     TransformingInstruction,
-    Variable,
+    Variable
 )
 
 _logger = logging.getLogger(__name__)
--- a/packages-nextgen/kestrel_core/src/kestrel/cache/sqlite.py
+++ b/packages-nextgen/kestrel_core/src/kestrel/cache/sqlite.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Iterable, Mapping, Optional
+from typing import Iterable, List, Mapping, Optional
 from uuid import UUID
 
 from dateutil.parser import parse as dt_parser
@@ -103,8 +103,8 @@ class SqliteCache(Cache):
         self,
         graph: IRGraphSoleInterface,
         instruction: Instruction,
-        stack: Optional[list[Instruction]] = None,
-    ) -> list[Instruction]:
+        stack: Optional[List[Instruction]] = None,
+    ) -> List[Instruction]:
         """
         Depth-first visit of nodes in graph from variable to variable or variable to source
 
--- a/packages-nextgen/kestrel_core/src/kestrel/cache/sqlite.py
+++ b/packages-nextgen/kestrel_core/src/kestrel/cache/sqlite.py
@@ -127,7 +127,7 @@ class SqliteCache(Cache):
         stack.append(instruction)
         return stack
 
-    def _evaluate_instruction_stack(self, stack: list[Instruction]) -> DataFrame:
+    def _evaluate_instruction_stack(self, stack: List[Instruction]) -> DataFrame:
         translator = SqliteTranslator()
         df = None
         previous_instruction = None
--- a/packages-nextgen/kestrel_core/src/kestrel/cache/sqlite.py
+++ b/packages-nextgen/kestrel_core/src/kestrel/cache/sqlite.py
@@ -17,7 +17,11 @@ from kestrel.ir.instructions import (
     Return,
     SourceInstruction,
     TransformingInstruction,
-    Variable
+    Variable,
+)
+
+from kestrel.exceptions import (
+    InevaluableInstruction,
 )
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


packages-nextgen/kestrel_core/src/kestrel/interface/datasource/codegen/sql.py
```diff
@@ -0,0 +1,156 @@
...
+def _render_comp(comp: FComparison) -> BinaryExpression:
+    col: ColumnClause = column(comp.field)
+    if comp.op == StrCompOp.NMATCHES:
+        return ~comp2func[comp.op](col, comp.value)
+    return comp2func[comp.op](col, comp.value)
+
+
+@typechecked
+def _render_multi_comp(comps: MultiComp) -> BooleanClauseList:
+    if comps.op == ExpOp.AND:
```
subbyte: Style comment (ignore if not like):
```
op = and_ if comps.op == ExpOp.AND else or_
```

packages-nextgen/kestrel_core/src/kestrel/interface/datasource/codegen/sql.py
```diff
@@ -0,0 +1,156 @@
...
+        op = and_
+    else:
+        op = or_
+    result = op(_render_comp(comps.comps[0]), _render_comp(comps.comps[1]))
```
subbyte: Style comment (ignore if not like):
```
result = reduce(op, map(_render_comp, comps.comps))
```

packages-nextgen/kestrel_core/src/kestrel/cache/sqlite.py
```diff
@@ -0,0 +1,151 @@
...
+        initial_cache: Optional[Mapping[UUID, DataFrame]] = None,
+        session_id: Optional[UUID] = None,
+    ):
+        super().__init__()
+
+        basename = self.session_id or "cache"
+        path = f"{basename}.db"
+
+        # for an absolute file path, the three slashes are followed by the absolute path
+        # for a relative path, it's also three slashes?
```
subbyte, line 51: I think so https://docs.sqlalchemy.org/en/14/core/engines.html#sqlite

Oh, it also supports SQLite in-memory database with two slashes, cool! No need to use the dataframe version of `InMemoryCache` then.

packages-nextgen/kestrel_core/src/kestrel/cache/sqlite.py
```diff
@@ -0,0 +1,151 @@
...
+            stack = []
+        # TODO: handle multiple predecessors of a node
+        _logger.debug("_visit: %s", instruction)
+        if isinstance(instruction, SourceInstruction) or (
+            isinstance(instruction, Variable) and len(stack) > 0
+        ):
+            # Need to first check if stack is empty since we start with var
+            # Since it's non-empty, this var should terminate the chain
+            stack.append(instruction)
+            return stack
```
subbyte: Style comment (ignore if not like):

Use `else:` for the following code block and only leave one `return` in this function. This may keep the control-flow simpler with one sink/return and reduce chance of error when adding code in the future.

packages-nextgen/kestrel_core/src/kestrel/cache/sqlite.py
```diff
@@ -0,0 +1,151 @@
...
+
+        stack = self._visit(graph, next(graph.predecessors(instruction)), stack)
```
subbyte: Should we raise an exception for invalid graph given to evaluate such as non-SourceInstruction as 0-indegree node?

```
Filter -> ProjAttr
```

`graph.predecessors(instruction)` will return empty and `next()` will hit a `StopIteration` in the next call.

cmadam (commented):
In `cache/sqlite.py` you are not using (because `TransformingInstruction`s are not handled yet ?) the `Filter` instruction.  If, however the code in `codegen/sql.py` handles the transforming instructions, you need to remove the `Filter` import.

In `codegen/sql.py`, the `render_exp()` function, do you need to also handle the NOT boolean operator?

packages-nextgen/kestrel_core/src/kestrel/cache/sqlite.py
```diff
@@ -0,0 +1,151 @@
...
+        return mapping
+
+    def _visit(
+        self,
+        graph: IRGraphSoleInterface,
+        instruction: Instruction,
+        stack: Optional[List[Instruction]] = None,
+    ) -> List[Instruction]:
+        """
+        Depth-first visit of nodes in graph from variable to variable or variable to source
```
subbyte: Do we assume the graph in input is a linear graph, so `next(graph.predecessors(instruction))` will cover all nodes backward?

1. If we have something like the following (`c` is a Filter node with reference to multiple variables):
```
a ->b; b->c; c->d; e->f; f->c

# this is coming from
a = NEW ...
e = NEW ...
d = a WHERE xxx = e.pid
```
We need to think about ways to deal with it.

2. If it is just a linear graph, why not call `translator.add_instruction(instruction)` from the sink node backward until any source node (`SourceInstruction` or `Variable`). This may result in combining `_visit()` and `_evaluate_instruction_stack` into one recursive function like:
```
def _evaluate_linear_instruction_graph(self, graph, instruction, translator):
    if `SourceInstruction` or `Variable` or `Ret`:
        ... # same as current _evaluate_instruction_stack()
    elif isinstance(instruction, TransformingInstruction):
        translator.add_instruction(instruction)
        _evaluate_linear_instruction_graph(self, graph, next(graph.predecessors(instruction)), translator)
```

subbyte (commented):
Nice to see the first SQL code! `sql.py` looks good in general, while I do not fully follow `sqlite.py` and have some questions.

PR #447, opened 2023-12-21
init session.py

subbyte:


cmadam (commented):
`packages-nextgen/kestrel_core/src/kestrel/session.py`: can you please add documentation with the definition for  the interface mapper?  Can we get a better name for the variable `self.im`?  Also, within a session can it be that more than one cache will be used?  In other words, would a more complicated structure be required ultimately for the interface mapper?

subbyte:
Good point! Added more information/comments on `session.im`. Usually we may have one cache per session, and we can allow users to specify which cache, e.g., `InMemoryCache` vs. `SqliteCache` to use in a future PR (when we have the config module).

subbyte:
Further revise the PR to put `name` attribute to each interface and simplify the interface locating in `session`.

PR #448, opened 2023-12-21
Entity name/attribute mapping to/from OCSF; use env vars in config

cmadam:
Perform the following: 
1. Map entity names and values to/from OCSF
2. Substitute environment variables defined in the configuration files with their values.  If the environment variables do not exist, the configuration will use their names as configuration values.


cmadam:
Following the review, the following changes were made:
1. Changes in the ECS to OCSF mappings of entity names and attributes
2. The method `from_ocsf_dictionaries()` maps each OCSF entity names or attributes to a list of values (of length 1 or larger)
3. Renamed the `src/kestrel/config/config.py` file to `src/kestrel/config/utils.py`
4. Marked the code in the `src/kestrel/utils.py` file that is handling backward compatibility with earlier versions of python and that will become deprecated using the function `is_python_older_than_minor_version()`
5. Added a test case (in `tests/test_config.py`) to illustrate the configuration behavior in case an environment variable specified in the configuration file has not been initialized.

subbyte (approved):
Thank you for your great work @cmadam !

Happy New Year!

PR #451, opened 2023-12-23
Use environment variables in configuration files

cmadam:
Provide capability to use environment variables in Kestrel configuration files (default or user-defined).  If the environment variable is defined, its value will be injected in the Kestrel configuration.  If the environment variable is not defined, its name will be injected in the Kestrel configuration.  For more details, see `tests/test_config.py`


subbyte (approved):
Well done! Thank you @cmadam !

PR #453, opened 2024-01-04
file prefetch improvements

subbyte:
- update default prefetch options
- improve file identifier

### DIFF
--- a/packages/kestrel_core/tests/test_command_find.py
+++ b/packages/kestrel_core/tests/test_command_find.py
@@ -85,24 +85,7 @@ def test_find_srcs_limit(set_empty_kestrel_config, fake_bundle_file):
         assert len(srcs) == 1
 
 
-def test_find_file_linked_to_process(set_empty_kestrel_config, proc_bundle_file):
-    with Session() as s:
-        stmt = f"""
-                procs = get process
-                        from file://{proc_bundle_file}
-                        where command_line LIKE 'wmic%'
-                files = FIND file LINKED procs
-                """
-        s.execute(stmt)
-        procs = s.get_variable("procs")
-        print(json.dumps(procs, indent=4))
-        assert len(procs) == 7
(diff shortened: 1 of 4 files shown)


packages/kestrel_core/tests/test_command_find.py
```diff
@@ -99,41 +99,42 @@ def test_find_file_linked_to_process(set_empty_kestrel_config, proc_bundle_file)
...
 
 def test_find_file_linked_to_process_limit_1(set_empty_kestrel_config, proc_bundle_file):
     with Session() as s:
         stmt = f"""
                 procs = get process
                         from file://{proc_bundle_file}
-                        where command_line LIKE 'wmic%'
-                files = FIND file LINKED procs LIMIT 1
+                        where name = 'svchost.exe'
+                ips = FIND ipv4-addr LINKED procs LIMIT 1
```
pcoccoli, line 94: The test name no longer matches what the test is doing.

subbyte: Good catch! Fixed.

pcoccoli: I meant that all these tests are named "find_file_linked_to_process" but now the all do `FIND ipv4-addr LINKED process`. So "find_ip_linked_to_process" is more appropriate.  By the way, do we have any tests left that do `FIND file LINKED process`?
I actually didn't notice the discrepancy with the limits.

subbyte: Yes, I updated the names to `find_ips_linked_to_process` to be consistent with `FIND ipv4-addr`.

In the previous commit, I changed some `FIND file` to `FIND ipv4-addr` to not affected by the prefetch default option change. There are still 3 tests with `FIND file`, which are not affected by the prefetch default option change, which I didn't update.

pcoccoli (approved):

PR #455, opened 2024-01-04
enable Python 3.12 in unit test

subbyte:


subbyte:
https://github.com/python/cpython/issues/114041

PR #459, opened 2024-01-15
K2 irg sourcehop

subbyte:
This is a preparation step for the Elasticsearch interface.


pcoccoli (approved):
It seems that OpenSearch (and therefore possibly ElasticSearch) does have some support for `JOIN`:  https://opensearch.org/docs/latest/search-plugins/sql/sql/complex/

PR #473, opened 2024-02-02
When resolving relations, also check for reflists

### LINKED ISSUE #472: directory CONTAINED file relationship not working
**Describe the bug**
`FIND directory CONTAINED x` (when x is a `file` variable) returns nothing but `FIND directory LINKED x` for the same `x` does return.

**Details of the bug**
The problem stems from relation to STIX property resolution when the STIX property is a reference list.  In this case, STIX defines a `directory:contains_refs` property.  For reference properties that are NOT lists (i.e. end in `_ref` and point to a single object rather than a list of `objects`), Kestrel will first check if the property exists in the current session.  For ref lists, this isn't done.

**To Reproduce**
Unit test to show the problem:
```
def test_find_dir_contained_file(set_empty_kestrel_config, proc_bundle_file):
    with Session() as s:
        stmt = f"""
                files = GET file
                        FROM file://{proc_bundle_file}
                        WHERE name = 'svchost.exe'
                dirs = FIND directory CONTAINED files
                """
        s.execute(stmt)
        files = s.get_variable("files")
        print(json.dumps(files, indent=4))
        dirs = s.get_variable("dirs")
        print(json.dumps(dirs, indent=4))
        assert len(dirs) == 1
```
This test will fail if the "contains_refs" property doesn't appear in the STIX bundle (it usually does not).  However, it should pass since `file:parent_directory_ref` is there.

pcoccoli:
Fixes #472

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,8 +18,9 @@ Fixed
 -----
 
 - Remove port checking in stix-shifter interface #452
-- Miss logging module in ``stix-shifter-diag``
-- Miss delete operation in ``kestrel_datasource_stixshifter`` #458
+- Add missing logging module import in ``stix-shifter-diag``
+- Add missing delete operation in ``kestrel_datasource_stixshifter`` #458
+- Fix relations like ``CONTAINS`` #472
 
 Changed
 -------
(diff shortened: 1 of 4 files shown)


subbyte (approved):
Look great!

Issue #269, opened 2022-12-07
Don't require schemes for `FROM` and `APPLY`


pcoccoli:
**Is your feature request related to a problem? Please describe.**
The hunter doesn't care if an analytic is using docker or python.

**Describe the solution you'd like**
`APPLY my_analytic` should work whether it's a python- or docker-based analytic.  Similarly `FROM my_datasource` should work without having to specify `stixshifter://my_datasource`.

Maybe we have something in config that lists interface preference order?  E.g. `python,docker` so it checks python first, then docker.  Stop at first match.

You could still supply the scheme, to force the one you want in case of name collisions.

**Describe alternatives you've considered**
Leave it like it is now.

**Additional context**
N/A

PR #474, opened 2024-02-06
kestrel2: opensearch datasource interface start

pcoccoli:


packages-nextgen/kestrel_core/src/kestrel/interface/manager.py
```diff
@@ -0,0 +1,94 @@
...
+    return list(itf_names)
+
+
+def _is_class(cls):
+    return lambda obj: inspect.isclass(obj) and obj.__bases__[0] == cls
+
+
+def _search_scheme_conflict(new_interface, interfaces):
+    for interface in interfaces:
+        for scheme_new in new_interface.schemes():
```
subbyte, line 90: We currently uses `.name()` instead of `.schemes()` in `AbstractDataSourceInterface` in Kestrel 2:
- Do we need to edit the method here to `.name()`?
- Do you think we need more comprehensive routing like Kestrel 1 (using the `InterfaceManager.schemes()` instead of the iterator of interfaces) to handle static dump data source such as a JSON file or parquet file with file:///, http://, and https://?

pcoccoli: Decoupling the scheme from the name might be useful for an interface that will handle multiple schemes, if we choose to do that.
Mostly I just copied code from Kestrel 1 and then hacked things up until it worked!

subbyte: I am towards not to support file dump data source for now, so we can keep this simple and move fast.

If so, we can remove the class `InterfaceManager`, rename `_load_interfaces()` into `load_interfaces()` and just call it in `session.py`.

pcoccoli: That's fine for datasources, but what happens when it's time to implement analytics interfaces?

subbyte: Analytics interfaces have name/scheme like `python://` or `awslambda://`, which should be fine. The additional consideration here (as you pointed out) is whether we allow any interface to be associated with multiple schemes, which should be OK not to support at this time I think.

I briefly think about analytics interface, which should be quite similar as our current datasource interface, if the graph segmentation method prepare the tiny subgraph for each analytics interface to evaluate.

I am absolutely fine if you'd like to keep it comprehensive, then we just use the dict instead of the iterator.

packages-nextgen/kestrel_datasource_opensearch/src/kestrel_datasource_opensearch/interface.py
```diff
@@ -0,0 +1,117 @@
...
+            else:
+                raise NotImplementedError(f"Unknown instruction type: {instruction}")
+
+        elif isinstance(instruction, SourceInstruction):
+            if isinstance(instruction, DataSource):
+                ds = self.config.indexes[instruction.datasource]
+                translator = OpenSearchTranslator(
+                    ds.timestamp_format, ds.timestamp, instruction.datasource
+                )
+                translator.add_instruction(instruction)
```
subbyte, line 113: Should we remove this line since the instruction is already infused in the constructor of `translator`? We may merge more information in the constructor if needed.

packages-nextgen/kestrel_datasource_opensearch/src/kestrel_datasource_opensearch/ossql.py
```diff
@@ -116,14 +117,18 @@ def _render_exp(self, exp: BoolExp) -> str:
             rhs = _render_comp(exp.rhs)
         return _and(lhs, rhs) if exp.op == ExpOp.AND else _or(lhs, rhs)
 
+    def add_DataSource(self, ds: DataSource) -> None:
```
subbyte, line 120: Suggest to remove this method and use the constructor to handle all table/index related info.

subbyte (approved):
Great work! You connect to the first real data source!

Just two things to consider before merge:
1. Whether we simplify DataSourceManager to just the load function (I can do it in followup PR if you like to focus making this interface work first in this PR).
2. Suggest to remove `add_DataSource` method and solely use the constructor of `OpenSearchTranslator` to pass in table/index information in the `FROM` clause.

packages-nextgen/kestrel_core/src/kestrel/interface/manager.py
pcoccoli: I would like to implement #269 in kestrel 2; that may require more advanced routing.

packages-nextgen/kestrel_datasource_opensearch/src/kestrel_datasource_opensearch/interface.py
pcoccoli: Sorry; I don't understand.  This is the only call to the constructor of `translator` and the `DataSource` instruction carries the name of the data source connection, which is needed here.

PR #497, opened 2024-04-16
example configs for kestrel2 interface modules

pcoccoli:


subbyte (commented):
Shall we add a comment at the beginning of each config about the names?

Maybe something like:
- `name` defined for each index/table will be used in Kestrel
- `name` defined in `connections` will be only used/referred inside the config

pcoccoli:
I added comments for the opensearch example; do you think this is a good format and sufficient info?

subbyte:
It is sufficient information, with one potential feature missing: name a data source, so that:
1. The huntbook does not need to be bound to the real index name from an OpenSearch instance.
2. Handle the case when two OpenSearch instances have the same index name.

A possible improvement may be:
```YAML
datasources:
  ds1:
    connection: opensearch1
    index: winlogbeat-*
    timestamp: "@timestamp"
    timestamp_format: "%Y-%m-%dT%H:%M:%S.%fZ"
    data_model_mapping: "custom_mapping.yaml"
  ds2:
    ...
  ds3:
    ...

connections:
  opensearch1:
    url: https://localhost:9200
    verify_cert: false
    auth:
      username: xxx
      password: yyy
  opensearch2:
    ...
```

pcoccoli:
Yes, separating the datasource name from the index name is a good idea.  I noticed that too while writing the example.  That will require some code changes - I'll start working on it.

PR #499, opened 2024-04-18
upgrade to stix-shifter v7 and hack verify_cert

subbyte:
1. upgrade to stix-shifter v7
2. though "disabling cert verification" is not supported in stix-shifter v7, it is still a very important feature, so I hacked stix-shifter to provide a Kestrel-only parameter in the `stixshifter.yaml` config to support it. The new parameter `verify_cert` is a boolean under `options` of `connection` (doc updated).


subbyte:
Good news! This stix-shifter v7 upgrade also fixes the issue for Python 3.12 (just tested locally), so we should be able to support Python 3.12 #455 after this upgrade.

packages/kestrel_datasource_stixshifter/src/kestrel_datasource_stixshifter/worker/utils.py
```diff
@@ -30,3 +32,19 @@ class TranslationResult:
...
+def disable_cert_verification_on_transmission(trans: StixTransmission):
+    ot = trans.entry_point.transmission()
+
+    # currently all the following attributes point to the same object
+    # iterate through them in case stix-shifter code changes in the future
+    for eps in [
+        "_BaseEntryPoint__ping_connector",
+        "_BaseEntryPoint__query_connector",
+        "_BaseEntryPoint__results_connector",
+        "_BaseEntryPoint__status_connector",
```
pcoccoli: Do we need the delete connector in here too?

subbyte: Good point! Rules generalized for enumerating the objects to patch.

pcoccoli (approved):

PR #508, opened 2024-04-26
Merge develop into develop_v1

pcoccoli:
Releases 1.8.3 and 1.8.4 were made from the develop branch, which is for v2.  We need to merge those changes into develop_v1 and then remove packages-nextgen from develop_v1 and rename it to packages in develop.


subbyte (approved):
Looks good to go!

PR #517, opened 2024-05-07
entity identifier config and multiple fixes

subbyte:
Add _entity identifier_ config to Kestrel 2, loaded to each datasource in each interface.

Besides that, did some cleanups and fixes on the config module and github actions:
1. fix `Makefile` for Kestrel 2
2. fix codecov github action for Kestrel 2
3. clean up `kestrel.yaml` default config for Kestrel 2
5. extend config loading to support `*.yaml` in config to load
6. remove the secondary loading code in interfaces
7. fix a Python 3.12 warning on `datetime`
8. add test cases for config loading

### DIFF
--- a/packages/kestrel_core/src/kestrel/config/kestrel.yaml
+++ b/packages/kestrel_core/src/kestrel/config/kestrel.yaml
@@ -1,70 +1,26 @@
 # syntax default values
 language:
-  default_variable: "_"
   default_sort_order: "desc"
-  default_datasource_schema: "stixshifter"
-  default_analytics_schema: "python"
-
-# how a Kestrel session is executed
-session:
-  cache_directory_prefix: "kestrel-session-" # under system temp directory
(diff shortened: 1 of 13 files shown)


packages/kestrel_core/src/kestrel/config/kestrel.yaml
```diff
@@ -1,70 +1,26 @@
...
+entity_identifier:
+  file: "hashes[?algorithm_id == 3]" # sha256
+  group: uid
+  process: uid
+  endpoint: ip
+  device: ip
+  src_endpoint:
+    - ip
+    - port
+  src_endpoint:
```
pcoccoli: Should be `dst_endpoint`

subbyte: Nice catch! Fixed.

pcoccoli (approved):

PR #519, opened 2024-05-08
frontend attr mapping

pcoccoli:


packages/kestrel_core/src/kestrel/frontend/compile.py
```diff
@@ -396,3 +405,39 @@ def explain(self, args):
...
+                stack.extend(graph.predecessors(curr))
+                if isinstance(curr, Variable):
+                    # Entity type HAS NOT been mapped yet for Variable
+                    native_type = curr.entity_type
+                    entity_type = self.entity_map.get(native_type, native_type)
+                elif isinstance(curr, ProjectEntity):
+                    # Entity type HAS been mapped for ProjectEntity
+                    native_type = curr.native_type
+                    entity_type = curr.entity_type
+                elif hasattr(curr, "attrs"):
```
subbyte: Is it better to explicitly use `isinstance(curr, ProjectAttrs)`?

If yes, we may also rename variable `has_attrs` to `ins_ProjectAttrs` to be specific.

packages/kestrel_core/src/kestrel/frontend/compile.py
```diff
@@ -396,3 +405,39 @@ def explain(self, args):
...
+            entity_type = None
+            has_attrs = []  # node with attrs that need to be mapped
+            while stack:
+                curr = stack.pop()
+                _logger.debug("curr = %s", curr)
+                stack.extend(graph.predecessors(curr))
+                if isinstance(curr, Variable):
+                    # Entity type HAS NOT been mapped yet for Variable
+                    native_type = curr.entity_type
+                    entity_type = self.entity_map.get(native_type, native_type)
```
subbyte: Shall we update `curr.entity_type` from its predecessors so we do not need to traverse to the beginning of the graph every time?

And we just break if both `native_type` and `entity_type` are not None; or we even do not need this traversal since we know the predecessor of a `ProjectAttrs` is directly a `Variable`, so we just walk one step to pull the info.

Maybe we could put "updating the `native_type` and `entity_type` of a node" as a method in `ir.graph.IRGraph`.

packages/kestrel_core/src/kestrel/ir/instructions.py
```diff
@@ -166,6 +169,7 @@ class Reference(IntermediateInstruction):
     """Referred Kestrel variable (used in AST) before de-referencing to a Kestrel variable"""
 
     name: str
+    entity_type: Optional[str] = None
```
subbyte: Sorry, could you remind me why do we need `entity_type` for `Reference`?

subbyte (commented):
Great to see the DISP attr mapping working!

Two things:
1. Do we want to support OCSF attributes in DISP? (see the additional test added)
2. I may have some ideas to improve the `_postprocessing()`, will discuss with you today (left some inlined notes here).

packages/kestrel_core/src/kestrel/frontend/compile.py
pcoccoli: I think I used `hasattr(curr, "attrs")` to possibly handle any other instructions.

packages/kestrel_core/src/kestrel/frontend/compile.py
pcoccoli: Update `curr.entity_type` to what?  This code is walking the graph in order to determine both the native and mapped entity types, which are needed for attribute mapping.

> we know the predecessor of a `ProjectAttrs` is directly a `Variable`

How do we know this?

> Maybe we could put "updating the `native_type` and `entity_type` of a node" as a method in `ir.graph.IRGraph`

That isn't what this code is doing; perhaps I need to improve the comments?

packages/kestrel_core/src/kestrel/ir/instructions.py
pcoccoli: I don't think it's needed; I can remove it.

packages/kestrel_core/tests/test_session.py
```diff
@@ -57,6 +57,27 @@ def test_execute_in_cache_stix_process():
...
+def test_execute_in_cache_stix_process_ocsf_disp_attr():
+    hf = """
+proclist = NEW process [ {"binary_ref.name": "cmd.exe", "pid": 123}
+                       , {"binary_ref.name": "explorer.exe", "pid": 99}
+                       , {"binary_ref.name": "firefox.exe", "pid": 201}
+                       , {"binary_ref.name": "chrome.exe", "pid": 205}
+                       ]
+DISP proclist ATTR name
+"""
+    b1 = DataFrame([ {"name": "cmd.exe"}
```
pcoccoli: Where is `name` supposed to come from here?

pcoccoli: `file.name` should work; `name` will not.  However, there is no attr mapping for `Construct` instructions at the moment, so even `file.name` won't work.  The cache has `binary_ref.name` as the column name.

packages/kestrel_core/src/kestrel/ir/instructions.py
```diff
@@ -120,6 +120,7 @@ def resolve_references(self, f: Callable[[ReferenceValue], Any]):
 @dataclass(eq=False)
 class ProjectEntity(SolePredecessorTransformingInstruction):
     entity_type: str
+    native_type: Optional[str]
```
subbyte: We may remove `Optional` since we always has `native_type` given when created a `ProjectEntity`

subbyte (approved):
Looks great!

PR #520, opened 2024-05-15
refactor SQL conditional translation

pcoccoli:
Make `_render_comp` and `_render_multi_comp` methods of `SqlTranslator` so that `SQLAlchemyTranslator` can use them.


subbyte (approved):
Looks good!

PR #522, opened 2024-06-06
make path transformers work for posix paths

pcoccoli:


subbyte (approved):
Nice!

To confirm I understand it right: now there are two mappings (Win/Linux) to translate a path from OCSF to native. Will both be in the native query compounded by `OR`?

pcoccoli:
Yes, that's the intent - I'm not sure I have a good test case for that, though.

PR #530, opened 2024-06-27
Kestrel 2 Python Analytics support

pcoccoli:
Also adds the CLI tools from kestrel 1.x

Some import changes due to running the commands below:
```
unimport --exclude __init__.py $SRC_DIRS
isort $SRC_DIRS
black $SRC_DIRS
```

`isort` re-ordered some imports

### DIFF
--- a/packages/kestrel_core/src/kestrel/analytics/interface.py
+++ b/packages/kestrel_core/src/kestrel/analytics/interface.py
@@ -1,0 +1,403 @@
+"""Python analytics interface executes Python function as Kestrel analytics.
+
+Use a Python Analytics
+----------------------
+
+Create a profile for each analytics in the python analytics interface config
+file (YAML):
+
+- Default path: ``~/.config/kestrel/pythonanalytics.yaml``.
+- A customized path specified in the environment variable ``KESTREL_PYTHON_ANALYTICS_CONFIG``.
+
+Example of the python analytics interface config file:
+
+.. code-block:: yaml
+
+    profiles:
+        analytics-name-1: # the analytics name to use in the APPLY command
+            module: /home/user/kestrel-analytics/analytics/piniponmap/analytics.py
+            func: analytics # the analytics function in the module to call
+        analytics-name-2:
+            module: /home/user/kestrel-analytics/analytics/suspiciousscoring/analytics.py
+            func: analytics
+
+Develop a Python Analytics
+--------------------------
+
+A Python analytics is a python function that follows the rules:
+
+#. The function takes in one or more Kestrel variable dumps in Pandas DataFrames.
+
+#. The return of the function is a tuple containing either or both:
+
+    - Updated variables. The number of variables can be either 0, e.g.,
+      visualization analytics, or the same number as input Kestrel variables.
+      The order of the updated variables should follow the same order as input
+      variables.
+
+    - An object to display, which can be any of the following types:
+
+        - Kestrel display object
+
+        - HTML element as a string
+
+        - Matplotlib figure (by default, Pandas DataFrame plots use this)
+
+   The display object can be either before or after updated variables. In other
+   words, if the input variables are ``var1``, ``var2``, and ``var3``, the
+   return of the analytics can be either of the following:
+
+   .. code-block:: python
+
+       # the analytics enriches variables without returning a display object
+       return var1_updated, var3_updated, var3_updated
+
+       # this is a visualization analytics and no variable updates
+       return display_obj
+
+       # the analytics does both variable updates and visualization
+       return var1_updated, var3_updated, var3_updated, display_obj
+
+       # the analytics does both variable updates and visualization
+       return display_obj, var1_updated, var3_updated, var3_updated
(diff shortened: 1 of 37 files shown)


packages/kestrel_core/src/kestrel/analytics/interface.py
```diff
@@ -0,0 +1,402 @@
...
+        return df
+
+
+class PythonAnalyticsInterface(AbstractInterface):
+    def __init__(
+        self,
+        serialized_cache_catalog: Optional[str] = None,
+        session_id: Optional[UUID] = None,
+    ):
+        _logger.debug("SQLAlchemyInterface: loading config")
```
subbyte: A typo?

packages/kestrel_core/src/kestrel/analytics/interface.py
```diff
@@ -0,0 +1,402 @@
...
+        if self.use_env:
+            os.environ = self.environ_original
+
+    def _execute(self, dataframe: DataFrame):
+        """Execute the analytics
+
+        Args:
+            dataframe (DataFrame): input variable to the analytics.
+
+        Returns:
```
subbyte, line 292: May update the docstring to avoid confusion.

packages/kestrel_core/src/kestrel/session.py
```diff
@@ -90,16 +95,19 @@ def evaluate_instruction(self, ins: Instruction) -> Display:
...
+                _logger.debug("eval: interface = %s", interface)
                 for iid, _display in (
                     interface.explain_graph(g)
                     if is_explain
-                    else interface.evaluate_graph(g)
+                    else interface.evaluate_graph(g, _cache)
                 ).items():
                     if is_explain:
                         display.graphlets.append(_display)
+                        if interface is not _cache:
```
subbyte: Shall we also remove the if statement like the evaluation case?

packages/kestrel_interface_opensearch/src/kestrel_interface_opensearch/interface.py
```diff
@@ -111,6 +109,7 @@ def store(
     def evaluate_graph(
         self,
         graph: IRGraphEvaluable,
+        cache: MutableMapping[UUID, Any],
```
subbyte, line 112: May create an issue (TODO) to use the cache.
`explain_graph()` also needs it.

packages/kestrel_interface_sqlalchemy/src/kestrel_interface_sqlalchemy/interface.py
```diff
@@ -191,6 +189,7 @@ def evaluate_graph(
     def explain_graph(
         self,
         graph: IRGraphEvaluable,
+        cache: MutableMapping[UUID, Any],
```
subbyte, line 192: May create an issue (TODO) to use the cache.
`explain_graph()` also needs it.

subbyte (commented):
Great work! Some minor in-line comments, plus a package structure update idea: does it make sense to rename `kestrel/analytics/interface.py` to `kestrel/analytics/python.py`?

packages/kestrel_core/src/kestrel/analytics/interface.py
pcoccoli: Yes - copy/paste error.  I will fix it.

packages/kestrel_core/src/kestrel/session.py
pcoccoli: I had to move this line here otherwise the explain unit tests failed - the `Construct` node would be omitted from the explain output, though I'm not completely certain why.

pcoccoli:
> does it make sense to rename kestrel/analytics/interface.py to kestrel/analytics/python.py?

It may, but my thinking here is that only the Python interface will be built-in (i.e. have its implementation in this package) while others such as docker will be implemented in their own packages (to avoid introducing unnecessary dependencies).

packages/kestrel_core/src/kestrel/session.py
subbyte: It is correct to omit the `Construct` node since it is already evaluated in the 3rd subgraph (`DISP domain`). The results of `domain` should be cached, preventing the 4th subgraph to trace back to the `Construct`.

Did a patch commit for it.

packages/kestrel_interface_opensearch/src/kestrel_interface_opensearch/interface.py
subbyte: https://github.com/opencybersecurityalliance/kestrel-lang/issues/532

packages/kestrel_interface_sqlalchemy/src/kestrel_interface_sqlalchemy/interface.py
subbyte: https://github.com/opencybersecurityalliance/kestrel-lang/issues/532

subbyte (approved):

PR #533, opened 2024-07-02
Upgrade Projection Entity

subbyte:
One critical dependent feature for `FIND`:
```
x = GET process:parent_ref FROM ... WHERE command_line = 'cmd.exe -abc'
```
Even we may not enable this in Kestrel syntax, the frontend should be able to add `ProjectEntity` node with such information to project `process:parent_ref` of a record, so we can get the parent info. If we only support project `process`, we will never be able to extract the parent process entity.

This requires two things:
1. Support `process.parent_process` as GET return/projection.
2. non-OCSF type/field in projection should be mapped to OCSF for internal data model consistency.
3. The actual return type should not be `process:parent_ref` (STIX) or `process.parent_process` (OCSF), both of which are fields, not type. The actual return type should be `process`.

## Solution

Re-organize the mappings into:
1. Field mapping
2. Projection mapping (a special type of field mapping like `process:parent_ref.*` mapped to `process.parent_process.*`
3. Projection field to type mapping

## What have been done

1. Rename/add mappings into two folders: `fields` and `types`
2. Add functionalities to support the use of projection mapping (in frontend)
3. Extend Kestrel syntax (Lark) to support `GET process:parent_ref`


packages/kestrel_core/src/kestrel/mapping/data_model.py
```diff
@@ -298,18 +304,35 @@ def translate_projection_to_native(
 
 
 @typechecked
-def translate_projection_to_ocsf(
-    dmm: dict,
+def translate_projection_entity_to_ocsf(
```
pcoccoli: This name seems a bit weird; maybe `translate_entity_projection_to_ocsf` would read better?

packages/kestrel_core/src/kestrel/mapping/data_model.py
```diff
@@ -298,18 +304,35 @@ def translate_projection_to_native(
...
+        native_projection += ".*"
+    ocsf_projection = _map.get(native_projection, native_projection)
+    if isinstance(ocsf_projection, list):
+        ocsf_projection = ocsf_projection[0]
+    ocsf_projection = ocsf_projection[:-2]
+    return ocsf_projection
+
+
+@typechecked
+def translate_projection_attrs_to_ocsf(
```
pcoccoli: Same as above; seems like it should be `translate_attr_projection_to_ocsf`

pcoccoli (approved):

packages/kestrel_core/src/kestrel/mapping/data_model.py
subbyte: Good point!

packages/kestrel_core/src/kestrel/mapping/data_model.py
subbyte: Good point!

PR #535, opened 2024-07-04
K2 FIND comamnd

### LINKED ISSUE #532: How to use the data from another interface/cache
When evaluating with multiple interfaces, we will store the results of one interface in cache, then give the reference of the cache to the next interface. The next interface will decide how to use the data from cache (e.g., a Kestrel variable) in its evaluation and explanation methods.

- If the interface supports WRITE, then we can create a new table for the incoming data, then do subquery or JOIN with the new table. Note that the data may be just a placeholder if it is in the explanation mode (virtual cache), and the previous results is not actually obtained.
- If the interface does not support WRITE (or not easily support WRITE), then we may enumerate the values from the cached variable. If there are too many items to enumerate, we may break the query into multiple smaller queries or raise an exception.

### LINKED ISSUE #534: Better way to handle variable type for APPLY
**Describe the bug**
Currently `self.variable_map` is used at `_KestrelT` to record the type of variable for later use. This is OK as long as the variable is defined in the same code block (IRGraph) as the APPLY. If a user only have one statement in the last code block: `APPLY ...`, then we only get a reference, but no variable. This will not work.

**Possible solution**
We may make this function into IRGraph merge(), so the variable type will be propagated to the variable after APPLY at that time (when two IRGraphs are merged).

subbyte:
## Done

1. Add relation resolution tables
2. Add multi-column support to RefComparison (matching both `pid` and `endpoint.id` for identifying `process`)
    - Update instruction and comparison data structure
    - Upgrade codegen for default SQL engine
    - Upgrade codegen for default Dataframe engine
    - Upgrade sql cache module to support multi-column in RefComparison
    - Upgrade inmemory cache module to support multi-column in RefComparison
4. Add CTE support to Sqlalchemy Interface
5. Add multi-column support of RefComparison to Sqlalchemy Interface
6. ~~Add `ReferenceDataSource` instruction~~ (will be resolved in parser directly)
7. Add `kestrel.irgraph.get_datasource_of_node()` method
8. Upgrade CTE support to Sqlalchemy Interface to fix a bug on retrieving dialect
9. Upgrade CTE support in sql cache module (applying the same code as above)
10. Add preliminary writable data lake in Sqlalchemy Interface support (resolved #532)
11. FIND parsing support (entity identifier config)
13. Add `irgraph` into `parse_kestrel()`
14. Fixed #534 
15. Upgrade all methods in parser transformer to update graph in place
16. Add transactional roll back in session if parsing failed
17. Add empty Filter support
18. IRGraph generation for FIND
19. Add support for ProjectEntity in SQL cache
20. Add tests
    - parser
    - cache inmemory (dataframe)
    - cache sql

## Todo (Next PR)

1. Add FIND tests for Sqlalchemy Interface


packages/kestrel_core/src/kestrel/interface/codegen/dataframe.py
```diff
@@ -12,10 +12,13 @@
     BoolExp,
     ExpOp,
     FExpression,
+    FBasicComparison,
+    RefComparison,
     ListOp,
     MultiComp,
     NumCompOp,
     StrCompOp,
+    AbsoluteTrue,
```
pcoccoli, line 21: Does `AbsoluteTrue` mean the `True` boolean literal?

subbyte: Yes, there is a chance we need empty Filter but with TimeRange:
```
x = GET event FROM schema://datalake WHERE ...
y = FIND process RESPONDED event
z = FIND process RESPONDED event START ... STOP ...
```
In these examples, `y` and `z` will have an empty `Filter` from `datalake` (a `DataSource` node) in case we need to append TimeRange (like in the case of `z`) before `ProjectEntity`.

pcoccoli (approved):

PR #543, opened 2024-07-15
Fixing EXPLAIN command

### LINKED ISSUE #541: EXPLAIN support
Need to test EXPLAIN after the big upgrade.

subbyte:
Add tests for EXPLAIN in both cache sql and sqlalchemy interface. Fixed #541 

Fixing the following issues:
1. Construct non-serializable due to the use of DataFrame (not supported by `mashumaro`)
2. `deepcopy()` will return `DataFrame` instead of our subclass
3. Outdated `explain_graph()` in sqlalchemy interface


packages/kestrel_core/src/kestrel/ir/instructions.py
```diff
@@ -33,6 +35,21 @@
     InitVar.__call__ = lambda *args: None
 
 
+class MashDataFrame(DataFrame, SerializableType):
```
pcoccoli: Not sure what `MashDataFrame` means.  If it's just a serializable DataFrame, maybe call it `SerializableDataFrame`?

packages/kestrel_core/src/kestrel/ir/graph.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import copy
```
pcoccoli, line 3: We should try to keep the import sorted according to PEP 8.  A simple way to do this is using the `isort` tool (I find i works best *after* `unimport` but *before* `black`).

packages/kestrel_core/src/kestrel/ir/instructions.py
subbyte: Good idea. Will rename

packages/kestrel_core/src/kestrel/ir/graph.py
subbyte: Thanks for the suggestion, will add to the github action.

pcoccoli (approved):

PR #545, opened 2024-07-18
data_model: allow literal value maps for transformers

pcoccoli:


packages/kestrel_core/src/kestrel/mapping/data_model.py
```diff
@@ -364,6 +389,29 @@ def translate_attributes_projection_to_ocsf(
...
+    return ocsf_value
+
+
+def _get_transformers(mappings: List[Union[str, dict]]) -> List[Union[str, dict]]:
+    """Collect value transformers from mappings, discarding duplicates"""
+    result = []  # Need a list instead of set since dicts aren't hashable
+    for mapping in mappings:
+        if "ocsf_value" in mapping or "native_value" in mapping:
+            transformer = _get_transformer(mapping)
+            if transformer not in result:
```
subbyte: Possible for `_get_transformer()` to return `None`? e.g. https://github.com/opencybersecurityalliance/kestrel-lang/blob/develop/packages/kestrel_core/src/kestrel/mapping/fields/ecs.yaml#L88

May not append `None` to `result`.

packages/kestrel_core/src/kestrel/mapping/data_model.py
```diff
@@ -374,17 +422,21 @@ def translate_dataframe(df: DataFrame, to_native_nested_map: dict) -> DataFrame:
...
                 else:
                     transformer_name = transformer_names.pop()
-                    df[col] = run_transformer_on_series(
-                        transformer_name, df[col].dropna()
-                    )
+                    if isinstance(transformer_name, dict):
+                        # Not actually a named function; it's a literal value map
+                        df[col] = df[col].replace(transformer_name)
+                    else:
+                        df[col] = run_transformer_on_series(
```
subbyte, line 438: May need a guard not to call `None` if it is passed back from `_get_transformers()`, or just not return `None` elements as in previous comment.

subbyte (commented):
Cool! Just a `None` type to check to ensure the cover all types from transformers.

packages/kestrel_core/src/kestrel/mapping/data_model.py
pcoccoli: Good catch

subbyte (approved):
Looks great!

PR #549, opened 2024-07-22
K2 multi datalake upgrade

subbyte:


packages/kestrel_interface_sqlalchemy/src/kestrel_interface_sqlalchemy/interface.py
```diff
@@ -60,6 +60,12 @@ def __init__(
     def schemes() -> Iterable[str]:
         return ["sqlalchemy"]
 
+    def get_storage_of_datasource(self, datasource: str) -> str:
+        """Get the storage name of a given datasource"""
+        if datasource not in self.config.datasources:
+            raise InvalidDataSource(datasource)
+        return self.config.datasources[datasource].connection
```
pcoccoli, line 67: This method is identical to the one in the opensearch interface.  Perhaps we should consider a `DatasourceInterface` derived from `AbstractInterface`?

pcoccoli (approved):

packages/kestrel_interface_sqlalchemy/src/kestrel_interface_sqlalchemy/interface.py
subbyte: Will introduce `DatasourceInterface` in the next PR for auto completion. However, since we do not template how to do config currently, I will leave this in each interface.

